PR #2, opened 2019-01-04
Feature/multiplebatteries

pkronstrom:


src/services/systemconfiguration.js
```diff
@@ -46,40 +44,56 @@ class SystemConfiguration {
         return matches;
     }
 
+    captureOne(regex, str) {
```
pkronstrom: This is not used, so should be removed

src/nodes/input-battery.html
```diff
@@ -21,37 +26,45 @@
...
+                    pathElem.val();
+            
+                    data.forEach(monitor => {
+                        serviceElem.append(
+                            $('<option/>', {value: JSON.stringify(monitor), text: monitor.name})
+                        );
+                    });
 
-                    let theBattery = data[Object.keys(data)[0]];
+                    let availablePaths = data[0]['paths'];
```
tkurki: Add a comment / TODO here?

src/services/systemconfiguration.js
```diff
@@ -46,40 +44,56 @@ class SystemConfiguration {
...
-
-            return services;
+        // parses all the dbus interfaces for batteries: com.victronenergy.battery.<id>
+        const re = /\bcom\.victronenergy\.battery\.(.*)/g;
+        
+        let dbusInterfaces = [...Object.keys(this.devices)];
+        let batteries = this.matchAndCapture(re, dbusInterfaces);
+        
+        let services = [];
+        batteries.forEach(battery => {
```
tkurki: `.map` would be clearer

src/services/victronclient.js
```diff
@@ -10,8 +10,8 @@ const debug = require('debug')('node-red-contrib-victron:victronclient')
  * both subscribe to dbus value updates as well as write values to dbus.
  */
 class VictronClient {
-    constructor(dbusAddress) {
-        this.dbusAddress = dbusAddress || '127.0.0.1'
+    constructor(dbusAddress, dbusPort) {
+        this.dbusAddress = dbusAddress || '127.0.0.1:78'
```
tkurki: Name shadow & confusion: is dbusAddress an ip address or addres + port?

src/nodes/config-client.js
```diff
@@ -4,6 +4,12 @@ module.exports = function(RED) {
 
     var VictronClient = require('../services/victronclient.js');
 
+    function serialize(key, value) {
```
tkurki: maybe `setReplacer` (per JSON.stringify documentation, there it is called `replacer`), as it is not really serializing

tkurki: pure function, so move to top level scope?

src/services/victronclient.js
pkronstrom: Indeed. I think it's better to have separate dbus address and port, but have only one environment variable to set when running node-red. I'll use separate address and port in VictronClient, and split the port and address from the environment variable upon initializing the config node.

src/nodes/config-client.js
pkronstrom: Good point, agreed.

src/nodes/input-battery.html
pkronstrom: The REST API (and the battery node as well) already supports for individual services per device. I will implement an onchange handler to change the measurement select to mirror the selected service.

Issue #13, opened 2019-01-10
Show paths per battery dynamically


pkronstrom:
At the moment, all battery monitors' services are the same static list of paths. Instead, show them dynamically, based on what paths are discovered in the cache.

Issue #6, opened 2019-01-08
Show CustomName in edit options


pkronstrom:
For example,
show `Battery 'House' - Voltage (V)` if a CustomName is set for the service, instead of the current `Battery - ttyO0 - Voltage (V)`

On the Backend-side, the VictronClient will need to make a separate request to /CustomName (or /ProductName) and register a handler for the incoming data. This has to be saved to the cache in order to show it on the Node-RED UI. I need to think of a good way to do this, so that it works on other devices, too.

---

Taken from a Slack discussion:

instead of the port, show the (custom) name of the device: use the text under /CustomName when available. And if not; then use /ProductName. The port (ie ttyO2), which is shown now, doesn't mean anything to installers. Herewith some examples:

```A solar charger, that doesn't even have a /CustomName path:

root@beaglebone:~# dbus -y com.victronenergy.solarcharger.ttyO2 /ProductName GetValue
'BlueSolar Charger MPPT 150/35 rev1'
root@beaglebone:~# dbus -y com.victronenergy.solarcharger.ttyO2 /CustomName GetValue
Traceback (most recent call last):
  File "/usr/bin/dbus", line 364, in <module>
    ret = obj.object.get_dbus_method(method.name, iface.name)(*args)
AttributeError: 'NoneType' object has no attribute 'name'

and a BMV, that does have a custom name:

root@beaglebone:~# dbus -y com.victronenergy.battery.ttyO0 /ProductName GetValue
'BMV-702'
root@beaglebone:~# dbus -y com.victronenergy.battery.ttyO0 /CustomName GetValue
'Hydraulic battery'
```

pkronstrom:
Current service & object-path discovery relies on dbus-listener.js polling the dbus root for available services. However, for e.g. battery monitors, this does not work properly, so the /ProductName is not visible to Node-RED.

See https://github.com/sbender9/signalk-venus-plugin/blob/master/dbus-listener.js#L102

Currently, I see a few options for this
a) fix the root polling somehow, so that all services support requesting the root path
b) implement a getValue function for dbus-listener.js to request specific values and initialize it on this 'device discovery' phase

Currently, the latter seems more plausible

mpvader:
Hey @pkronstrom; something must be wrong there: all service you care about support a GetValue() or GetText() method call on the root path (`/`).

Try on a real system. Perhaps the issue is that the dbus-recorder doesn't support this? Although I think that was also fixed. Not sure though.

[...]

Issue #5, opened 2019-01-08
Group all nodes under Victron Energy category


pkronstrom:
Simpy add `category: 'Victron Energy'` in registerType function in html

pkronstrom:
<img width="203" alt="screenshot 2019-01-09 at 16 10 14" src="https://user-images.githubusercontent.com/11190027/50904490-16307580-1429-11e9-8ff5-72aaed25302e.png">

Issue #4, opened 2019-01-08
Use a Victron Energy icon on nodes


pkronstrom:
Requires a properly formatted icon. Easy to add after that.

From node-red documentation: https://nodered.org/docs/creating-nodes/appearance
"A node can provide its own icon in a directory called icons alongside its .js and .html files. These directories get added to the search path when the editor looks for a given icon filename. Because of this, the icon filename must be unique.

**The icon should be white on a transparent background, 20 x 30 in size.**"

pkronstrom:
Fitting VE logo to a 20px wide png (as instructed in the documentation) makes it look quite fuzzy. Trying SVG instead.

<img width="654" alt="screenshot 2019-01-09 at 10 59 54" src="https://user-images.githubusercontent.com/11190027/50888760-0a2fbe00-13ff-11e9-9cf9-bd6c7fa19e0f.png">

PR #16, opened 2019-01-10
Victron node enhancements

pkronstrom:
Includes issues #3, #4, #5, #6 and #13


src/services/victronclient.js
```diff
@@ -65,13 +70,43 @@ class VictronClient {
...
+     * @param {object} msg a message object received from the dbus-listener
+     */
+    saveToCache(msg) {
+        //let dbusInterface = this.cache[msg.senderName] || {}
+        let dbusInterface = {}
+        if (this.system.cache[msg.senderName]) {
+            dbusInterface = this.system.cache[msg.senderName]
+        } else {
+            // Upon first discovery, request the customname and productname
+            // of the battery. The returned message gets saved to the cache.
```
tkurki, line 94: Do we get notified of customName updates somehow?

src/services/victronclient.js
```diff
@@ -16,8 +16,12 @@ class VictronClient {
     constructor(address, port) {
         this.dbusAddress = address || '127.0.0.1'
         this.dbusPort = port || '78'
+
         this.subscriptions = {} // an array of subscription objects [{ "topic": topic, "handler": function }, ...]
+
         this.write
+        this.request
```
tkurki, line 23: How about naming these two `this._getValue` and `this._setValue`, underscore to point out they are for _internal use_? This naming obfuscates what these are doing.

tkurki: Is there a possibility that something tries to use these before they are initialised? I guess not, especially when we get to using dbus directly and the connect should never fail.

src/services/victronclient.js
pkronstrom: An update to the customName should sends an update message to the dbus, iirc. I need to test this with real hardware when possible, though.

src/services/victronclient.js
pkronstrom: Good idea, both are internal, and setValue is exposed via VictronClient.publish() anyways.

pkronstrom: In theory, the relay node could try publishing before the dbus-listener is initialized, good catch. I feel this should be guarded somehow: maybe check that the value of this.connection before writing to dbus.

PR #61, opened 2020-05-13
feature: support all data points in the system node

sbender9:


mpvader:
Note that I just now made an issue for all or part of the above work; and only afterwards found this PR.

Here is the issue:

https://github.com/victronenergy/node-red-contrib-victron/issues/84

src/services/services.json
```diff
@@ -1437,43 +1439,402 @@
...
+                "path": "/Buzzer/State",
+                "type": "float",
+                "name": "Buzzer State",
+                "enum": {
+                    "0": "Off",
+                    "1": "On"
+                }
+            },
+            {
+                "path": "/SystemState/State",
```
dirkjanfaber: We don't want to add /SystemState here, as those are not exposed via modbus_tcp.

dirkjanfaber (changes requested):
We don't want to expose /SystemState for now, as that is also not exposed via modbus_tcp. We could consider generating  the `services.json` file from the excel file in the https://github.com/victronenergy/dbus_modbustcp repository.
Either that, or add a filter that checks the connection type and exposes only the services for the used connection. I'll make a new issue for that.

PR #155, opened 2022-12-16
Added "ESS -> Max charge power" to node options

mr-manuel:
I added the "ESS -> Max charge power" since I saw in the community forum that also others could need it.

https://community.victronenergy.com/questions/128814/how-to-control-charge-current-of-multiplus-with-no.html?childToView=175112


dirkjanfaber:
Looks ok to me, only a remark needs to be added that it is not used with DVCC. But I'll do that.

dirkjanfaber (approved):

PR #162, opened 2023-02-22
changed: fixed some descriptions

mr-manuel:
On some descriptions the word "phase" was missing.


dirkjanfaber (approved):
LGTM. Though it will also require to regenerate the documentation of the nodes and add it to the wiki. I'll do that. (see point 4 on https://github.com/victronenergy/node-red-contrib-victron#adding-new-nodes-developers)

PR #168, opened 2023-03-09
Include gridmeter /Ac/Power as "Power (W)"

GallusMax:
..otherwise the L1+L2+L3 Power values had to be summed up The /Ac/Power value is already offered in service-whitelist.js, ready for use.

### DIFF
--- a/src/services/services.json
+++ b/src/services/services.json
@@ -2204,6 +2204,11 @@
         "name": "Total Reverse Energy (sold) (kWh)"
       },
       {
+        "path": "/Ac/Power",
+        "type": "float",
+        "name": "Power (W)"
+      },
+      {
         "path": "/Ac/L1/Current",
         "type": "float",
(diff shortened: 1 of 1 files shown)


dirkjanfaber:
Note the `service-whitelist.js` file is not used anymore (at least not by me), so that is not so relevant. The nodes take the values from the dbus and if the data is not there, adding it to `src/nodes/services.json` won't make any difference.

For this case, there is an example added to the nodes (via _import (ctrl-i) -> examples tab ->  @vicrtronenergy/node-red-contrib-victron_): *Sum power inputs*

image

This example can also be downloaded from [here](https://raw.githubusercontent.com/victronenergy/node-red-contrib-victron/master/examples/Sum%20power%20inputs.json). The other examples are also [here](https://github.com/victronenergy/node-red-contrib-victron/tree/master/examples).

If this is not sufficient, please create a new issue for it.

PR #199, opened 2024-10-15
Add Relay-Option to SolarCharger Output

patrick-dmxc:
Add Relay-Option to SolarCharger Output
Needed, to control the Reley of e.g.
SmartSolar 450 MPPT

### DIFF
--- a/scripts/service-whitelist.js
+++ b/scripts/service-whitelist.js
@@ -461,7 +461,8 @@ module.exports = {
   'output-solarcharger': {
     solarcharger: [
       '/Mode',
-      '/MppOperationMode'
+      '/MppOperationMode',
+      '/Relay/0/State'
     ]
   },
   'output-ess': {


patrick-dmxc:
Hi @dirkjanfaber,
can you take a look on this?
I think this should work

dirkjanfaber:
I am afraid this won't work. The `services-whitelist.js` file isn't used at all by the node code. All services are defined in the `services.json` file. Documentation is then generated from that, using the `service2doc.js` script.

But I am also fairly sure that controlling the relays is already functional for the solar charger. See https://www.victronenergy.com/live/venus-os:large#controlling_relays for more information on that.

patrick-dmxc:
But its not showing up in node-red while its set to Manuel in the settings of the MPPT 450, so it should work, but it diddn't
image

dirkjanfaber:
You need to use the Relay control node instead of the Solar Charger Control node.

patrick-dmxc:
This is working also not working
Its just showing the RElays of the Cerbo
image

dirkjanfaber:
I will need to check this on an actual system. Hope to get to that this Friday.

patrick-dmxc:
Thanks

patrick-dmxc:
Shomething new here?

dirkjanfaber:
Sorry, forgot about this. Bit found a system with a comparable MPPT now and able to reproduce it. 
Will get in the next release.

PR #210, opened 2025-01-14
issue: Is not reflected in VRM UI

podarok:
image


podarok:
image

/Connected is not in dbus-spy nor in the list of Measurements in NodeRED

podarok:
I see, it is fixed in master https://github.com/victronenergy/node-red-contrib-victron/commit/73252a8938ecf5d6d8a49b00f8c5b485d539cfb9

PR #211, opened 2025-01-15
fix incorrect port number

MikeP-NZ:
The instructions incorrectly use port 1880 instead of 1881


dirkjanfaber:
Well, there was an error there. Besides changing the port, it also requires to change the protocol (to https). Adjusted that in the instructions.

PR #212, opened 2025-01-15
feat: Add ability to assign single phase virtual PV inverter to any of L1-L3 phases

podarok:
image
image
image

### DIFF
--- a/src/nodes/victron-virtual.js
+++ b/src/nodes/victron-virtual.js
@@ -280,6 +280,10 @@ module.exports = function (RED) {
             iface['Dc/0/Temperature'] = 25
             iface.Soc = 80
           }
+          if (!config.include_battery_temperature) {
+            delete ifaceDesc.properties['Dc/0/Temperature']
+            delete iface['Dc/0/Temperature']
+          }
           text = `Virtual ${properties.battery.Capacity.format(iface.Capacity)} battery`
           break
         }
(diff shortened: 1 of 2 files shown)


src/nodes/victron-virtual.js
```diff
@@ -346,7 +352,7 @@ module.exports = function (RED) {
             iface.ErrorCode = 0
             iface.StatusCode = 0
           }
-          text = `Virtual ${iface.NrOfPhases}-phase pvinverter`
+          text = `Virtual ${iface.NrOfPhases}-phase pvinverter on phase ${iface.SinglePhaseNr}`
```
podarok: this needs work. If 3 phase - then text should be different

podarok:
The only issue I see not fixed - in gui-v2
image

I can't find where this is hardcoded for single phase PV

dirkjanfaber:
Continued this PR in [virtual-pvinverter-single-phase-select](https://github.com/victronenergy/node-red-contrib-victron/tree/virtual-pvinverter-single-phase-select)

That number on the gui is a bit weird. What I can find is https://github.com/victronenergy/gui-v2/blob/dd07063999a82acaa0844f0a56d4fe3e6a2f230f/pages/settings/devicelist/ac-in/PageAcInModel.qml#L25, which I've implemented. What I notice is, when I set `/NrOfPhases` to 2, I am getting the L2 (but also a L1). 

image

podarok:
image
works good when 3 phase PV inverter 
Not working in gui-v2 when single phase on L3
image

PR #214, opened 2025-02-13
Null values support

Chris927:
This enables setting a value to null, by using the Custom Control node.

Note that, other then the small change in this package, the latest versions of some dependencies are required for this to work.

I've left a comment on how we might provide better feedback (in case of high latency or faults). This is not related to getting null values to work, though.

### DIFF
--- a/src/nodes/victron-nodes.js
+++ b/src/nodes/victron-nodes.js
@@ -155,6 +155,22 @@ module.exports = function (RED) {
         }
 
         if (!this.pathObj.disabled && this.service && writepath) {
+
+          // If the value is null, just call. (experimental)
+          // Note: it's not ideal that we set the status before the call to dbus (via this.client.publish()) has happened.
+          // Error handling and feedback might be better, if we wait for the dbus call to return, and then only
+          // set the status. Or better even, set the status to something like "busy..." or "working..." while the
+          // dbus call is happening.
+          if (value === null) {
+            this.client.publish(this.service, writepath, value)
+            this.node.status({
+              fill: 'green',
+              shape,
+              text: 'set to null'
+            })
+            return
+          }
+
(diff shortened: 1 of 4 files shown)


src/nodes/victron-nodes.js
```diff
@@ -155,6 +155,22 @@ module.exports = function (RED) {
...
+          // Note: it's not ideal that we set the status before the call to dbus (via this.client.publish()) has happened.
+          // Error handling and feedback might be better, if we wait for the dbus call to return, and then only
+          // set the status. Or better even, set the status to something like "busy..." or "working..." while the
+          // dbus call is happening.
+          if (value === null) {
+            this.client.publish(this.service, writepath, value)
+            this.node.status({
+              fill: 'green',
+              shape,
+              text: 'set to null'
```
dirkjanfaber, line 169: Prefer to have that with a capital S. Other status texts usually start with a capital letter too.

src/nodes/victron-nodes.js
```diff
@@ -155,6 +155,22 @@ module.exports = function (RED) {
         }
 
         if (!this.pathObj.disabled && this.service && writepath) {
+
+          // If the value is null, just call. (experimental)
+          // Note: it's not ideal that we set the status before the call to dbus (via this.client.publish()) has happened.
+          // Error handling and feedback might be better, if we wait for the dbus call to return, and then only
+          // set the status. Or better even, set the status to something like "busy..." or "working..." while the
+          // dbus call is happening.
```
dirkjanfaber, line 163: I agree that it would probably be better to update the status when we get the feedback from the dbus call, but that would also mean that we need to track the node that does the call (so we can update the status of the correct node).

dirkjanfaber (approved):
Nice work, @Chris927.

PR #220, opened 2025-03-12
Better feedback on progress and success or failure when setting a dbus value

Chris927:
Previously, when setting a value via `this.client.publish(...)`, we set the status without knowing if the value was actually set successfully,.

Now, we use the callback provided to only set "success!" once done. We also handle errors, and display the error in the status, if there is an error message available.


Chris927:
Watch out! When testing locally, with a flow from an earlier version, I get the following warnings on the console, and the flow has unknown node types:

```
12 Mar 18:00:58 - [info] Waiting for missing types to be registered:
log.js:91
12 Mar 18:00:58 - [info]  - victron-client
log.js:91
12 Mar 18:00:58 - [info]  - victron-virtual
log.js:91
12 Mar 18:00:58 - [info]  - victron-output-custom
log.js:91
12 Mar 18:00:58 - [info]  - victron-input-temperature
```

I suspect there is something else required before this can be merged :)

dirkjanfaber (approved):
Looks good to me. Thanks!

PR #226, opened 2025-03-27
Add missing TemperatureTypes to Virtual Temp Devices

patrick-dmxc:
Hi, while testing the 3.60 beta i found that the Virtual Temp-Devices did not support other Types then Battery, Fridge and Generic.
So is added the other Types.


dirkjanfaber (approved):
LGTM. Thanks for adding that, @patrick-dmxc

Issue #219, opened 2025-03-05
[Feature] poll less frequently (or stop polling)


dirkjanfaber:
**Describe the bug**

Every 5 seconds the poller queries _all_ info from the dbus, causing an extra load of the system. This has been in the code for several years already, so apparently it hasn't been a real issue. But it would be wise to handle this in a better way. 

Relevant code is here: https://github.com/victronenergy/node-red-contrib-victron/blob/master/src/services/dbus-listener.js#L117

Chris927:
TLDR; I did some performance profiling, and this indicates that the polling is the cause for the majority of the CPU work done by the node-red process.

Disclaimer: As is common with performance profiling, results depend on the environment and configuration of the system under test; given a different environment or another configuration, these results may not be reproducible.

The citations below are from the file produced by `node --prof-process isolate-0xnnnnnnnnnnnn-v8.log > processed.txt`, see [the NodeJS docs](https://nodejs.org/en/learn/getting-started/profiling) for more information on profiling.

The profiling run was done on a Cerbo GX, with node-red running for a few hours, without any user interaction, and one connected device (a Ruuvi temperature sensor).

## Where is time spent?

The summary shows that 85.7% are spent in shared libraries.

```
 [Summary]:
   ticks  total  nonlib   name
  194953   14.3%  100.0%  JavaScript
      0    0.0%    0.0%  C++
  294258   21.7%  150.9%  GC
  1164176   85.7%          Shared libraries
```

Under "Shared libraries", 65.4% (889,501 ticks) are spent in node. (I don't understand why node is considered a shared lib here, but that should not matter in terms of the result.)

It is worth noting that 21.7% spent on garbage collection ("GC") indicate some level of memory pressure. (This might not be a problem, yet, though.)

```
 [Shared libraries]:
   ticks  total  nonlib   name
  889501   65.4%          /data/home/root/.nvm/versions/node/v22.14.0/bin/node
  259978   19.1%          /lib/libc.so.6
   5695    0.4%          /usr/lib/libstdc++.so.6.0.32
   4584    0.3%          /lib/libgcc_s.so.1
   4155    0.3%          [vdso]
    261    0.0%          /lib/libm.so.6
      2    0.0%          /lib/ld-linux-armhf.so.3
```

[...]

Issue #185, opened 2023-09-08
[BUG] Uncontrolled/unexpected node msg reception


pippajb:
I have noticed with recent past versions and and have just confirmed using v3.10~36 that there are issues with the rate of message reception from the victron nodes.

I am finding that the '5 second' transmission rate does not work, for example I count messages received for gps speed and see a continuous stream rather than one every 5 seconds (this did work sometime back).

Another 'related' issue is with multiplus switch position, here again the 5 second rule doesn't work, but further 'on change' doesn't either. Again I see multiple messages coming in even when there is no actual switch change (and the payload is unchanged from the previous)! I believe there are other nodes at fault also.

The result of this issue is heavily loading my dashboard.

**Steps to reproduce the behaviour**

This is a snapshot of my test flow:
![image](https://github.com/victronenergy/node-red-contrib-victron/assets/144439684/48cbd4ae-a981-4d4d-9b70-72208e26d248)

Both counts increase unexpectedly.

**Hardware**
 -  Venus GX

**Versions**
 - node-red-contrib-victron version: 1.5.0
 - Venus OS version: v3.10~36

dirkjanfaber:
Thank you. I'll try to reproduce this. Which add-on is used for the counter?

pippajb:
Hi, the counter is just a simple node to increment a global when triggered, one global for each Count., the count being showed as a status.

dirkjanfaber:
Understood, but which add-on provided that? I now used [node-red-contrib-msg-counter](https://flows.nodered.org/node/node-red-contrib-msg-counter) and that one seems to count as I would have expected.

pippajb:
I've  just tested using node-red-contrib-msg-counter and still see the same issue, odd.

My flow is as follows:
_[
    {
        "id": "afd84ad68566263c",
        "type": "subflow",
        "name": "Subflow 1",
        "info": "",
        "category": "",
        "in": [
            {
                "x": 60,
                "y": 80,
                "wires": [
                    {
                        "id": "4bc84f12b60f5f8a"
                    }
                ]
            }
        ],
        "out": [
            {
                "x": 320,
                "y": 80,
                "wires": [
                    {
                        "id": "4bc84f12b60f5f8a",
                        "port": 0
                    }
                ]
            }
        ],
        "env": [
            {
                "name": "Global",
                "type": "str",
                "value": ""
            }
        ],
        "meta": {},
        "color": "#DDAA99",
        "status": {
[...]

PR #230, opened 2025-04-23
Remove polling

Chris927:
Introduces an environment variable `ENABLE_POLLING`, so that polling (compare #219, #185) can optionally be replaced with periodically calling subscribers with data from the cache of dbus properties we maintain.

Unless you set `ENABLE_POLLING=false`, the behavior in terms of polling is as before, so *not polling* is opt-in for now. At a later stage, default behavior can be flipped to not poll by default.

More testing is required. Specifically, I haven't tested yet that a dbus service joining the bus will be detected and cached with its properties correctly. @dirkjanfaber , I'd appreciate if you can test this :)

Once we are happy with this, it should resolve related performance concerns, as per #219.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
             "version": "1.6.24",
             "license": "MIT",
             "dependencies": {
-                "dbus-native-victron": "^0.4.3",
+                "dbus-native-victron": "^0.4.4",
                 "dbus-victron-virtual": "^0.1.12",
                 "debug": "^4.4.0",
                 "lodash": "^4.17.21",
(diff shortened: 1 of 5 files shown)


dirkjanfaber:
@Chris927 This works neatly. I've modified the branch a bit (https://github.com/victronenergy/node-red-contrib-victron/commit/6743d1929b1afce3f5100a466dc80b2bc708633a) with another commit, changing from an environment variable to a configuration option. Making the whole a bit more user friendly.

Chris927:
> @Chris927 This works neatly. I've modified the branch a bit ([0373131](https://github.com/victronenergy/node-red-contrib-victron/commit/037313183b7d4badb328f204fbc9322f645766f7)) with another commit, changing from an environment variable to a configuration option. Making the whole a bit more user friendly.

Excellent, thank you!

PR #237, opened 2025-05-30
Virtual node reconnect dbus

Chris927:
A virtual node can lose connection to dbus, as it is a TCP connection. Here, we ensure we re-connect each virtual node in case this happens:

- We listen for 'connect', 'end' and 'error' events on the dbus connection.
- If we lost connection, we will retry connecting again after a delay of 1 second.

### DIFF
--- a/src/nodes/config-client.html
+++ b/src/nodes/config-client.html
@@ -1931,7 +1931,7 @@ Questions can also be asked on the Victron Energy community:
   <li>11 - Monthly</li>
   <li>-1 - Disabled</li>
 </ul>
-<p>A negative value means that the schedule has been de-activated.</p></dd>
+</dd>
 
   <dt class="optional">Schedule 1: Duration (seconds)<span class="property-type">integer</span></dt>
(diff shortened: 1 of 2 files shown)


src/nodes/victron-virtual.js
```diff
@@ -600,30 +600,62 @@ module.exports = function (RED) {
...
+          'Could not connect to the DBus session bus.'
+        )
+        node.status({
+          color: 'red',
+          shape: 'dot',
+          text: 'Could not initialize DBus.'
+        })
+        return
+      }
+      // TODO: we are not really connected yet
```
dirkjanfaber: That TODO can probably go, same as the one below.

src/nodes/victron-virtual.js
```diff
@@ -978,7 +1011,7 @@ module.exports = function (RED) {
             return { loaded: false }
           }
 
-          // console.warn(`Loading saved data for ${config.device}: ${JSON.stringify(savedData)}`)
+          console.warn(`Loading saved data for ${config.device}: ${JSON.stringify(savedData)}`)
```
dirkjanfaber: This can become a `debug`.

PR #243, opened 2025-06-17
Virtual device persistence

Chris927:
Virtual devices have state, but other than real devices, their state disappears when node-red restarts. This PR adds persistence of selected state of virtual devices into the file system.

Currently, the flag `persist` in the `properties` table of victron-virtual.js controls if and how often to persist:

- If the flag is not present, the value is not persisted. On restart, the property gets its default value.
- If the flag is set to `true`, the value gets persisted whenever the value changes. This is the correct setting for e.g. a switch state, to ensure we don't lose the state.
- If the flag is set to a number, then the value is persisted, but not written more often than every N seconds, whereas N is the value of the `persist` flag.

The location of the persisted state in the file system is `/data/home/nodered/.victron` by default, but can be adjusted either through settings (see `.vscode/settings.js`) or through an environment variable `PERSISTED_STATE_LOCATION`.

### DIFF
--- a/src/nodes/persist.js
+++ b/src/nodes/persist.js
@@ -105,15 +105,59 @@ function setupToWriteLater (RED, id, iface, ifaceDesc, propName) {
   }
 }
 
+const savedStateCache = {} // Cache for saved state to avoid redundant writes, indexed by id
+
+function isSaveNeeded (id, iface, ifaceDesc, propName) {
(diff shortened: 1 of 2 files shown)


Chris927:
We want one squashed commit, not many small, intermediate commits. Therefore I have created [a new PR](https://github.com/victronenergy/node-red-contrib-victron/pull/2440) which supersedes this one

PR #252, opened 2025-07-10
Setting values on virtual devices

Chris927:
Virtual nodes can now receive input, so that properties of virtual devices can be updated.

Example: Use an "inject" node to set the payload to `{ "Capacity":122, "Level":23.73 }` to update properties for a virtual tank. Alternatively, the property names can start with `/`, e.g. `/Capacity` instead of `Capacity`.

This also fixes an unrelated bug: Previously, after a re-deploy, `VictronDbusListener` might have stopped responding to `ItemsChanged` signals, see the comment in `node.on('close', ...`.

### DIFF
--- a/src/services/dbus-listener.js
+++ b/src/services/dbus-listener.js
@@ -399,7 +399,7 @@ class VictronDbusListener {
     if (typeof value === 'string') { numType = 's' }
 
     try {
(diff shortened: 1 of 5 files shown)


src/services/dbus-listener.js
dirkjanfaber, line 1: I'd rather have this as a `debug` instead.

PR #256, opened 2025-07-20
fix: add wildcard expansion for relays

Chris927:
`SystemConfiguration.getRelayServices()` requires wildcard expansion, as the new services.json uses wildcards also for system relays. This fix adds wildcard expansion to `getRelayServices()`.

We also try/catch to avoid errors without anything in the logs.

### DIFF
--- a/src/services/victron-system.js
+++ b/src/services/victron-system.js
@@ -81,12 +81,14 @@ class SystemConfiguration {
       const pathObjects = paths.map(p => {
         const svc = service.split('.')[2].split('/')[0] // com.victronenergy.system => system
 
-        const relayObject = _.find(_.get(utils.SERVICES, ['relay', svc]), { path: p })
-
-        if (!relayObject) { console.error(`A relay path specification '${p}' is missing in services.json for service ${svc}.`) }
-
+        // iterate over utils.SERVICES.relay[svc], and expand
+        try {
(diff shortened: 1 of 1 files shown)


mpvader:
thanks! So to make sure I understand this correctly: this regression is not because of a change in Venus OS / dbus-systemcalc wrt relay handling, but because recent changes in node-red-contrib-victron itself?

That would be good, then I don't have to worry about any of the other integrations like Modbus-TCP and such.

dirkjanfaber:
Correct. This regression comes from [this change](https://github.com/victronenergy/node-red-contrib-victron/commit/5e5a158d0550e4d84d4065f21d4e9a2df454dba7) where the `services.json` file was reduced in size by introducing wildcard expansion to that file. I have overlooked the effects for the relays (which have this seperate logic).

dirkjanfaber (approved):
The changes look good to me. Only thing that I would like to see added is a test for the relay to the `test/wildcard-expansion.test.js` file.
I've added that in [this commit](https://github.com/victronenergy/node-red-contrib-victron/commit/ff7e59582fcc0788ffb92c5f54cb0654a06971b5) (which  also could use some verifcation).

mpvader:
I'll merge this and get it on the todo; the extra test case can come later @dirkjanfaber - thanks!

PR #263, opened 2025-08-12
dbus: fix stream write error handling, fix scenario where we don't call `setValue` if not connected yet

Chris927:
The actual fix is in dbus-native, where the callback passed to `invoke()` [is now forwarded](https://github.com/Chris927/dbus-native/commit/bc514dbc544d8f45da3ffd92fd5d89f5feaf7d7f#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R129-R131) to the call to `stream.write()`.

Previously, there was no callback passed to stream.write() at all. If an error occurred while writing to the stream, that error did not bubble up to the caller. Now it does.

To fix, we upgrade to `dbus-native-victron` v0.4.5.

We also remove a check on `client.connected`: The check is unnecessary, as the dbus client buffers messages until it is connected. The check is also harmful, as we won't retry if we didn't call `setValue()` while not connected yet.

### DIFF
--- a/src/services/victron-client.js
+++ b/src/services/victron-client.js
@@ -72,6 +72,15 @@ class VictronClient {
       }
     }
 
+    // Support dependency injection for testing
+    if (options.dbusClient) {
+      debug('Using injected dbus client for testing')
+      this.client = options.dbusClient
+      // For mocked clients, mark as connected
+      this.client.connected = true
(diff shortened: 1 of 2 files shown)


dirkjanfaber (approved):
Tested and everything works as expected. Thank you!

PR #264, opened 2025-08-15
fix: Do not set unexpected virtual battery voltage

podarok:
Reported-in: https://victrondevelopment.slack.com/archives/C04RYU0UMDE/p1755246359776809


Chris927 (approved):

dirkjanfaber:
Before merging, this PR needs some changes (which I'll do one of these days). The default value might not be ideal. Prefer to have that selectable via a dropdown (12, 24 or 48 V or custom or so).

Issue #266, opened 2025-08-19
Grid meter frequency values rounded to 0.05 Hz


dirkjanfaber:
**Describe the bug**

Via https://community.victronenergy.com/t/grid-meter-frequency-values-rounded-to-0-05-hz-lost-precision-compared-to-earlier-behavior/42280

> I’m using Node-RED on a Venus OS system to monitor the grid frequency via a Grid Meter (com.victronenergy.grid/31, path /Ac/Frequency).
> I’ve noticed that the frequency values now only change in 0.05 Hz steps, for example:
> `49.95 → 50.00 → 50.05 → 50.00 → ... `
> Previously, these values were provided with higher resolution, such as 49.97 or 50.02 Hz.

This is via the grid meter input node and an EM540 energy meter.

dirkjanfaber:
Closing this as completed (after I got confirmation of the user reporting it).It is part of 1.6.43.

PR #270, opened 2025-08-20
Fix dbus message processing for devices without Text fields + add unit tests

### LINKED ISSUE #267: Grid meter isn't updating
Via https://community.victronenergy.com/t/v3-70-node-red-grid-meter-isn-t-updating/40564

> In node red the grid meter frequency isn’t updating.
> I only get the latest frequency after deploy or restart flows.

Also using a EM540.

dirkjanfaber:
Issue #267: Some dbus devices (like EM540) send `ItemsChanged` messages with only `Value` fields, missing the `Text` representation. The original code required both `Value` AND `Text` fields (`entry[1].length === 2`), causing these messages to be ignored.

**Example of problematic message:**
```json
[
  "/Ac/Frequency",
  [
    [
      "Value",
      [
        [{"type": "d", "child": []}],
        [49.971000000000004]
      ]
    ]
  ]
]
```

**Example of working message:**
```json
[
  "/Ac/L1/Power",
  [
    ["Text", [[{"type": "s", "child": []}], ["1332W"]]],
    ["Value", [[{"type": "d", "child": []}], [1331.7]]]
  ]
]
```

### Solution
1. **Fix the filtering logic**: Changed from `entry[1].length === 2` to `entry[1]` to accept messages with any number of fields
2. **Remove Text requirement**: Changed filtering from `!m.path || m.value === null || !m.text` to `!m.path || m.value === null`

### Refactoring for Testability  
Extracted the dbus message processing logic into `src/services/core/dbus-message-processor.js` to enable proper unit testing of this complex parsing behavior.

- ✅ **Backwards compatible** - existing functionality unchanged
- ✅ **Unit tested** - both Value-only and Value+Text scenarios covered  
- ✅ **Maintainable** - complex parsing logic now isolated and testable
- ✅ **Standard compliant** - follows project code style guidelines

### Testing
```bash
npm test -- tests/core.dbus-message-processor.test.js
```

Tests cover:
- Messages with both Value and Text fields (existing case)
- Messages with Value only (issue #267 fix)
- Complete message processing with mixed entry types
- Proper filtering of null values

Fixes #267 _and_ #266 (confirmed by the user)

### Backporting
Not deeming this as a backport candidate as ignoring the Value-only dbus messages has been the default behaviour until now.

### DIFF
--- a/src/services/core/dbus-message-processor.js
+++ b/src/services/core/dbus-message-processor.js
@@ -1,0 +1,60 @@
+function parseItemsChangedEntry (entry) {
+  const m = { changed: true }
+  m.path = entry[0]
+
+  if (entry[1]) {
+    entry[1].forEach(v => {
+      switch (v[0]) {
+        case 'Value':
+          m.value = v[1][1][0]
+          break
+        case 'Text':
+          m.text = v[1][1][0]
+          break
+      }
+    })
+  }
+
+  return m
+}
+
+function processItemsChanged (msg, services, searchDeviceInstanceByName) {
+  const messages = []
+
+  if (!msg.body || !msg.body[0]) {
+    return messages
+  }
+
+  msg.body[0].forEach(entry => {
+    const m = parseItemsChangedEntry(entry)
+
+    if (!m.path || m.value === null) {
+      return
+    }
+
+    const service = services[msg.sender]
+    if (!service || !service.name) {
+      return
+    }
+
+    if (m.path === '/DeviceInstance') {
+      service.deviceInstance = m.value
+    }
+
+    m.senderName = service.name.split('.').splice(0, 3).join('.')
+
+    if (service.deviceInstance === null) {
+      service.deviceInstance = searchDeviceInstanceByName(services, m.senderName, '')
+    }
+
+    m.deviceInstance = service.deviceInstance
+    messages.push(m)
+  })
+
+  return messages
+}
+
+module.exports = {
+  parseItemsChangedEntry,
+  processItemsChanged
+}
--- a/test/core.dbus-message-processor.test.js
+++ b/test/core.dbus-message-processor.test.js
@@ -1,0 +1,94 @@
+const {
+  parseItemsChangedEntry,
+  processItemsChanged
+} = require('../src/services/core/dbus-message-processor')
+
+describe('dbus message processing - issue #267 fix', () => {
+  const entryWithBothFields = [
+    "/Ac/L1/Power",
+    [
+      [
+        "Text",
+        [
+          [{ "type": "s", "child": [] }],
+          ["1332W"]
+        ]
+      ],
+      [
+        "Value", 
+        [
+          [{ "type": "d", "child": [] }],
+          [1331.7]
+        ]
+      ]
+    ]
+  ]
+  
+  const entryWithValueOnly = [
+    "/Ac/Frequency",
+    [
+      [
+        "Value",
+        [
+          [{ "type": "d", "child": [] }],
+          [49.971000000000004]
+        ]
+      ]
+    ]
+  ]
+
+  test('parses entry with both Value and Text', () => {
+    const result = parseItemsChangedEntry(entryWithBothFields)
+    
+    expect(result.path).toBe('/Ac/L1/Power')
+    expect(result.value).toBe(1331.7)
+    expect(result.text).toBe('1332W')
+    expect(result.changed).toBe(true)
+  })
+
+  test('parses entry with Value only (the fix for issue #267)', () => {
+    const result = parseItemsChangedEntry(entryWithValueOnly)
+    
+    expect(result.path).toBe('/Ac/Frequency')
+    expect(result.value).toBe(49.971000000000004)
+    expect(result.text).toBeUndefined()
+    expect(result.changed).toBe(true)
+  })
+
+  test('processes complete message with mixed entry types', () => {
+    const msg = {
+      sender: 'com.victronenergy.vebus.ttyUSB0',
+      body: [[entryWithBothFields, entryWithValueOnly]]
+    }
+    
+    const services = {
+      'com.victronenergy.vebus.ttyUSB0': {
+        name: 'com.victronenergy.vebus.ttyUSB0',
+        deviceInstance: 0
+      }
+    }
+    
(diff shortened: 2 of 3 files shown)


src/services/core/dbus-message-processor.js
```diff
@@ -0,0 +1,60 @@
+function parseItemsChangedEntry (entry) {
+  const m = { changed: true }
+  m.path = entry[0]
```
Chris927, line 3: There are some scenarios possible here which we would silently ignore, example:

If `entry` happens to be a string (it shouldn't, but could), then we would assign the first character of that string to `m.path`.

This is particularly nasty / dangerous, because NodeJS is dynamically typed, so we have no idea (at this point of the code) what the value of `entry` is: it could be a boolean, a string, and array (that's what we want), whatever.

The approach I suggest going forward: validate as strictly as possible. Here, we could do this:

```diff
diff --git a/src/services/core/dbus-message-processor.js b/src/services/core/dbus-message-processor.js
index c9c68be..4a7a017 100644
--- a/src/services/core/dbus-message-processor.js
+++ b/src/services/core/dbus-message-processor.js
@@ -1,4 +1,11 @@
+function validateItemsChangedEntry(entry) {
+  if (!Array.isArray(entry) || entry.length < 1) {
+    throw new Error('Invalid entry format, expected an array with at least one element')
+  }
+}
+
 function parseItemsChangedEntry(entry) {
+  validateItemsChangedEntry(entry)
   const m = { changed: true }
   m.path = entry[0]

diff --git a/test/core.dbus-message-processor.test.js b/test/core.dbus-message-processor.test.js
index 2b2ab12..b405ec1 100644
--- a/test/core.dbus-message-processor.test.js
+++ b/test/core.dbus-message-processor.test.js
@@ -46,6 +46,11 @@ describe('dbus message processing - issue #267 fix', () => {
     expect(result.changed).toBe(true)
   })

+  test('parsing fails if not an array with at least one element', () => {
+    expect(() => parseItemsChangedEntry('not an array')).toThrow('Invalid entry format, expected an array with at least one element')
+    expect(() => parseItemsChangedEntry([])).toThrow('Invalid entry format, expected an array with at least one element')
+  })
+
   test('parses entry with Value only (the fix for issue #267)', () => {
     const result = parseItemsChangedEntry(entryWithValueOnly)
```

src/services/core/dbus-message-processor.js
```diff
@@ -0,0 +1,60 @@
...
+  const messages = []
+
+  if (!msg.body || !msg.body[0]) {
+    return messages
+  }
+
+  msg.body[0].forEach(entry => {
+    const m = parseItemsChangedEntry(entry)
+
+    if (!m.path || m.value === null) {
```
Chris927, line 31: Why do we not do anything in this case? Maybe explain in a comment.

Is it worthwhile logging a warning if this happens?

src/services/core/dbus-message-processor.js
```diff
@@ -0,0 +1,60 @@
...
+      return
+    }
+
+    const service = services[msg.sender]
+    if (!service || !service.name) {
```
Chris927, line 36: If we don't find the service, it's probably fine to just ignore the message. But maybe we should have a debug log in that case?

If the service has no name, is this a bug of sorts? Is it worth a `console.warn()`?

src/services/core/dbus-message-processor.js
```diff
@@ -0,0 +1,60 @@
...
+      return
+    }
+
+    if (m.path === '/DeviceInstance') {
+      service.deviceInstance = m.value
```
Chris927, line 41: `service` is passed to us as an argument (via `services`), now we change a value in there. That's a side effect. Is this intended?

I see we did this in the old code, too. I suggest adding a comment like `// TODO: this is a side effect, we may not want this, avoid once we have better test coverage`.

Once we have better test coverage (unit testing *and* integration testing) we can try and remove those side effect. As long as we have them, they may introduce hard-to-debug issues.

src/services/core/dbus-message-processor.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    m.senderName = service.name.split('.').splice(0, 3).join('.')
```
Chris927, line 44: Maybe this is nothing we should do in this PR, but we should refactor all those splits and splices and joins we do into helper functions that do it for us. We do `split('.').splice(0, 3).join('.')` in 8 locations in the code at the moment.

Maybe add a function with a name that indicates why we do this (I actually don't know why): The service names I've seen all had two `.` in them, so `split('.').splice(0, 3).join('.')` is a noop.

src/services/core/dbus-message-processor.js
```diff
@@ -0,0 +1,60 @@
...
+
+    if (service.deviceInstance === null) {
+      service.deviceInstance = searchDeviceInstanceByName(services, m.senderName, '')
```
Chris927, line 47: Same as above, the assignment of `service.deviceInstance` is a side effect. Do we want this?

test/core.dbus-message-processor.test.js
```diff
@@ -0,0 +1,94 @@
+const {
+  parseItemsChangedEntry,
+  processItemsChanged
+} = require('../src/services/core/dbus-message-processor')
+
+describe('dbus message processing - issue #267 fix', () => {
```
Chris927, line 6: It's so awesome to finally have those unit tests! Yay!

test/core.dbus-message-processor.test.js
```diff
@@ -0,0 +1,94 @@
...
+        [
+          [{ "type": "d", "child": [] }],
+          [49.971000000000004]
+        ]
+      ]
+    ]
+  ]
+
+  test('parses entry with both Value and Text', () => {
+    const result = parseItemsChangedEntry(entryWithBothFields)
```
Chris927, line 41: We could have unit tests for `parseItemsChangedEntry` in their own `describe() { ... }` block. `describe`s can be nested: Generally, we could have one `describe('...') {...}` for each function (and even more `describe()` blocks to separate out scenarios per function, if needed, see [here](https://tinytip.co/tips/testing-nested-describe/) for an example.

Chris927 (approved):
We should follow up on some of my comments.

In terms of the fix, we can merge it, though. Therefore I'm approving this PR now.

PR #280, opened 2025-09-15
feat: configurable default voltage for virtual battery

dirkjanfaber:
Adds 12V/24V/48V/custom presets with fallback to 24V. Also add the TimeToGo path and updates documentation.

Realising that the test itself isn't that good.

Supersedes PR #274. That one was depending on a large re-factor. Which is still going to happen, but might as wel do this one first.

Also adds the missing path `/TimeToGo` (requested by @philipptrenz)


src/nodes/victron-virtual.js
```diff
@@ -546,8 +547,25 @@ module.exports = function (RED) {
               iface.Capacity = Number(config.battery_capacity)
             }
             if (config.default_values) {
+              let voltage = 24 // fallback
+
+              if (config.battery_voltage_preset === '12') {
```
Chris927: Can be made less repetitive:

```javascript
if (['12', '24', '48'].includes(config.battery_voltage_preset)) {
  voltage = Number(config.battery_voltage_preset)
} else if (config.battery_voltage_preset === 'custom') {
  ...
}
```

Maybe also extract the array into a variable:

```javascript
const BATTERY_DEFAULT_VOLTAGES = ['12', '24', '48']`;
```

This avoids hard-coding "magic numbers" in functions. We do not need this global variable anywhere else (yet), but it is cleaner to keep magic numbers out of functions, if possible.

src/nodes/victron-virtual.js
```diff
@@ -82,6 +82,7 @@ const properties = {
     'Info/ChargeRequest': { type: 'i', format: (v) => v != null ? v : '', value: 0 },
     Soc: { type: 'd', min: 0, max: 100, format: (v) => v != null ? v.toFixed(0) + '%' : '', persist: 15 /* persist, but throttled to 15 seconds */ },
     Soh: { type: 'd', min: 0, max: 100, format: (v) => v != null ? v.toFixed(0) + '%' : '', persist: 60 /* persist, but throttled to 60 seconds */ },
+    TimeToGo: { type: 'd', max: 86400, format: (v) => v != null ? v.toFixed(0) + 's' : '' },
```
Chris927, line 85: Is this change intentional? It seems unrelated to the PR.

If intentional, mention it in the PR description.

Chris927 (commented):
When trying to test it, I get `404 Not Found` for http://localhost:1880/resources/@victronenergy/node-red-contrib-victron/victron-virtual-functions.js, even though I ran `npm run build`. What am I missing?

Should file `resources/victron-virtual-functions.js` still be under source control? My understanding is that it gets generated from `src/nodes/victron-virtual-functions.js` now.

src/nodes/victron-virtual.js
dirkjanfaber: It was indeed intentional. Updated the PR.

src/nodes/victron-virtual.js
dirkjanfaber: Great suggestions. I've updated the code via [this](a1906365cfb58cca27ecb5a1e65a7d2bb17eafc2) commit.

dirkjanfaber:
> Should file resources/victron-virtual-functions.js still be under source control? My understanding is that it gets generated from src/nodes/victron-virtual-functions.js now.

Yes, that is a good observation too. I did ad it to `.gitignore`, bot overlooked removing it from git. Dit that [now](9ba6bd64f1dc42f1285275b6f79a3e8423f50d05).

PR #281, opened 2025-09-17
fix(dbus-listener): improve error handling for GPS service and virtual reconnects

dirkjanfaber:
- Move console.warn outside virtual service check to log all reconnection attempts
- Add specific handling for com.victronenergy.gps.ve_ttyUSB2 GetItems errors
- Prevent GPS service errors from breaking dbus listener initialization


Chris927:
We are working around an issue in the GPS service implementation, can we raise an issue there? Eventually the issue should be fixed there, and our special handling should become unnecessary.

PR #282, opened 2025-09-17
feat(victron-virtual): add UI control visibility for virtual switches

dirkjanfaber:
Just a minor thing I noticed the gui-v2 complaining about. 
This field controls the visibility in gui-v2. Keeping it persistent to make sure that, if the value is set, it gets remembered.

### DIFF
--- a/src/nodes/victron-virtual.js
+++ b/src/nodes/victron-virtual.js
@@ -800,7 +800,8 @@ module.exports = function (RED) {
                 }[v] || 'unknown'),
                 persist: false
               },
-              { name: 'Settings/ValidTypes', type: 'i', value: 0x7 }
+              { name: 'Settings/ValidTypes', type: 'i', value: 0x7 },
+              { name: 'Settings/ShowUIControl', type: 'i', value: 1, persist: true }
             ]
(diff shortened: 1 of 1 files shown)


src/nodes/victron-virtual.js
```diff
@@ -800,7 +800,8 @@ module.exports = function (RED) {
                 }[v] || 'unknown'),
                 persist: false
               },
-              { name: 'Settings/ValidTypes', type: 'i', value: 0x7 }
+              { name: 'Settings/ValidTypes', type: 'i', value: 0x7 },
+              { name: 'Settings/ShowUIControl', type: 'i', value: 1, persist: true }
```
Chris927, line 804: To confirm my understanding: This makes a setting `ShowUIControl` available, but this setting is not configurable through the Node-RED interface, correct?

I'm approving, assuming that this is the desired behavior :)

Chris927 (approved):

src/nodes/victron-virtual.js
dirkjanfaber: That is correct.

PR #287, opened 2025-09-19
Default customname of switch to the swich type

dirkjanfaber:
Instead of setting the Name of the switch to "Switch [1-4]", set it to reflect the type of the switch. E.g. "Temperature setpoint".


Chris927 (approved):
Looks good, tested it.

I thought about having a unit test for this change, but as we are within the `proceed()` function, this would be really hard before the refactor. Hope we'll get the refactor in soon :)

Issue #292, opened 2025-09-23
Switches: Set ValidTypes path correctly


ReinvdZee:
Switches created in NR all set the `/ValidTypes` path to 7, indicating that it can be of type momentary, toggle or dimmable. This is also the case for switches which are not of type momentary, toggle or dimmable. This causes the GUI to display the type selection menu as unknown:

<img width="1902" height="519" alt="Image" src="https://github.com/user-attachments/assets/24a28e9e-ef08-4ca4-8cca-6a949202f9e5" />

Changing the type on the go often times also requires some paths to be added. I.e., when switching from toggle from dimmable, the `/Dimming` path must be added if it's not already there. This currently is allowed (because `/ValidTypes == 7`), but does not work, as `/Dimming` is not added to the service.

Since the type is already configurable in NR, it is the easiest to not support changing the switch type from the GUI. The solution is to set `/ValidTypes` to `1 >> type` which will cause the GUI to display just the type which it currently has. The dropdown menu won't do anything for the user, but it does make it clear that no other type can be selected.

PR #297, opened 2025-09-26
Set ValidTypes path correctly

dirkjanfaber:
Set /ValidTypes to 1 >> type which will cause the GUI to display just the type which it currently has.

See issue #292

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -46,15 +46,15 @@
     //     }
     //   ]
     // },
-    // 7: {
-    //   label: 'Basic slider',
-    //   fields: [
-    //     { id: 'min', type: 'number', placeholder: 'Min value', title: 'Slider minimum', style: 'width:80px;' },
-    //     { id: 'max', type: 'number', placeholder: 'Max value', title: 'Slider maximum', style: 'width:80px;' },
-    //     { id: 'step', type: 'number', placeholder: 'Step size', title: 'Step size', style: 'width:80px;' },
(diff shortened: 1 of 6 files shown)


src/nodes/victron-virtual.js
```diff
@@ -824,6 +824,10 @@ module.exports = function (RED) {
                 if (name === 'Settings/Type') propValue = switchType
 
                 iface[key] = propValue !== undefined ? propValue : 0
+
+                if (name === 'Settings/ValidTypes') {
+                  iface[key] = 1 << Number(config[`switch_${i}_type`])
```
Chris927: This is confusing, as `Number(config[`switch_${i}_type`])` is almost the same as `switchType`, except for when it is null or undefined. Is this intentional?

Could this code be improved by making some variable names more specific? `key` is a generic name, maybe something like `switchOutputPropertyKey`?

I don't understand why bit-shifting is the right thing here. Maybe a comment or reference to some place that explains this would help.

Chris927 (changes requested):
This code is hard to read. Some suggestions on how it can possibly be improved with little effort:

1.

src/nodes/victron-virtual.js
dirkjanfaber: Fair points. Should be resolved by [this](https://github.com/victronenergy/node-red-contrib-victron/pull/297/commits/252b656e26446683d65768ae4c0ebdc88865bb57) commit.

Chris927 (approved):

PR #299, opened 2025-09-29
Virtual switch: add dropdown

dirkjanfaber:
Select from a custom list of options defined by you. The selected value is stored in the `Dimming` field (starting at `0` for the first option).

Note for Chris: Initial design was to have key / pair value stored as json. Now we just store the values in an array and have the selected one stored in the Dimming field.

### DIFF
--- a/src/nodes/victron-virtual.html
+++ b/src/nodes/victron-virtual.html
@@ -593,6 +593,7 @@ monitored remotely. There are currently seven types of switches available:
 - **Temperature setpoint**: Sets a target temperature value
 - **Stepped switch** : Select up to 7 discrete steps (e.g., fan speeds)
 - **Dropdown**: Select from a custom list of options defined by you. The selected value is stored in the `Dimming` field (starting at `0` for the first option).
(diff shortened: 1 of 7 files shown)


src/nodes/victron-virtual.html
```diff
@@ -127,22 +127,19 @@
...
-                                    keyValueObj[key] = value;
+                                if (value) {
+                                    labelsArray.push(value);
                                 }
                             }
-                            
-                            // Store as simple key-value JSON object
-                            this[`switch_${i}_label`] = JSON.stringify(keyValueObj);
+                            // Store as JSON string array
+                            console.log(labelsArray)
```
Chris927: Either remove log statement, or make output clearer (e.g. `console.log("dropdown labels:", labelAsArray)`).

dirkjanfaber: Resolved in [this](https://github.com/victronenergy/node-red-contrib-victron/pull/299/commits/2dc686831c6411f175cf380e7658d4b1a71cdae5) commit.

Issue #240, opened 2025-06-11
Docs: add Virtual Devices


mpvader:
- A chapter to our node-red manual
- examples, both hardcopy in the wiki as well as inside the flow editor, if they are not there already

PR #303, opened 2025-10-02
Reduce the nr of switches from 4 to just 1

dirkjanfaber:
For maintainability and future expansion.
See https://community.victronenergy.com/t/venus-os-v3-70-40-available-for-public-testing/45708/56 for more context.

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -84,26 +84,26 @@
     }
   };
 
-  function renderSwitchConfigRow (i, context) {
+  function renderSwitchConfigRow (context) {
     const typeOptions = Object.entries(SWITCH_TYPE_CONFIGS)
       .map(([value, cfg]) => `<option value="${value}">${cfg.label}</option>`)
       .join('');
     const switchRow = $(`
         <div class="form-row">
-            <label for="node-input-switch_${i}_type"><i class="fa fa-toggle-on"></i> Switch ${i} type</label>
-            <select id="node-input-switch_${i}_type" data-switch-index="${i}">${typeOptions}</select>
+            <label for="node-input-switch_1_type"><i class="fa fa-toggle-on"></i> Switch 1 type</label>
+            <select id="node-input-switch_1_type">${typeOptions}</select>
         </div>
     `);
     $('#switch-config-container').append(switchRow);
 
-    const savedType = context[`switch_${i}_type`] !== undefined ? context[`switch_${i}_type`] : 1;
-    $(`#node-input-switch_${i}_type`).val(String(savedType));
+    const savedType = context.switch_1_type !== undefined ? context.switch_1_type : 1;
+    $('#node-input-switch_1_type').val(String(savedType));
 
     function renderTypeConfig () {
-      $(`#switch-${i}-config-row`).remove();
-      $(`#switch-${i}-pairs-row`).remove();
+      $('#switch-1-config-row').remove();
+      $('#switch-1-pairs-row').remove();
 
-      const type = $(`#node-input-switch_${i}_type`).val();
+      const type = $('#node-input-switch_1_type').val();
       const cfg = SWITCH_TYPE_CONFIGS[type];
 
       if (cfg && cfg.fields.length) {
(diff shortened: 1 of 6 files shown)


resources/victron-virtual-functions.js
```diff
@@ -84,26 +84,26 @@
...
-  function renderSwitchConfigRow (i, context) {
+  function renderSwitchConfigRow (context) {
     const typeOptions = Object.entries(SWITCH_TYPE_CONFIGS)
       .map(([value, cfg]) => `<option value="${value}">${cfg.label}</option>`)
       .join('');
     const switchRow = $(`
         <div class="form-row">
-            <label for="node-input-switch_${i}_type"><i class="fa fa-toggle-on"></i> Switch ${i} type</label>
-            <select id="node-input-switch_${i}_type" data-switch-index="${i}">${typeOptions}</select>
+            <label for="node-input-switch_1_type"><i class="fa fa-toggle-on"></i> Switch 1 type</label>
```
Chris927, line 93: I assume you keep the `_1` part of the input id to be backwards-compatible, i.e. by not changing the config (field names) for the first switch of the node, the first switch should still work as configured, just switches 2, ..., 4 won't work (they will disappear, I guess).

But: The name of the label can safely be "Switch type" (without the "1"), can it not? "Switch 1" is likely going to be confusing for users who never saw the ability to have multiple switches per node (and this is likely going to be the majority).

src/nodes/victron-virtual.js
```diff
@@ -804,231 +802,229 @@ module.exports = function (RED) {
...
 
-              // Set CustomName and Group from config
-              const customNameKey = `SwitchableOutput/output_${i}/Settings/CustomName`
-              ifaceDesc.properties[customNameKey] = {
-                type: 's',
+            // Set CustomName and Group from config
+            const customNameKey = 'SwitchableOutput/output_1/Settings/CustomName'
+            ifaceDesc.properties[customNameKey] = {
+              type: 's',
+              persist: true
```
Chris927, line 838: As discussed: It appears that `persist: true` practically makes the config property un-editable. On the UI, it seems you can update, but it has no effect on the property; the issue can be confirmed via `dbus-spy`.

See [this PR](https://github.com/victronenergy/node-red-contrib-victron/pull/306) to fix this.

src/nodes/victron-virtual.js
```diff
@@ -804,231 +802,229 @@ module.exports = function (RED) {
...
 
-                // DimmingMin
-                const minKey = `SwitchableOutput/output_${i}/Settings/DimmingMin`
-                ifaceDesc.properties[minKey] = {
-                  type: 'd',
-                  format: (v) => v != null ? v.toFixed(1) + '°C' : ''
-                }
-                iface[minKey] = Number(config[`switch_${i}_min`] ?? 0)
+              // StepSize
+              const stepKey = 'SwitchableOutput/output_1/Settings/StepSize'
```
Chris927, line 891: There is an issue with the "source of truth" of the StepSize: The StepSize is stored both in the configuration (in the flow) and in dbus, and both values are writable, and they are not getting synced from dbus to the configuration.

The UI uses the step size available via dbus.

This means I can change the step size via dbus, which will reflect in the UI. However, when I deploy the flow again, the step size will be overwritten from the flow.

We cannot easily sync from dbus to the flow, so that is not a solution.

When opening the configuration, can we read from dbus at that moment? This may also not be easy.

I'm not sure what a good solution is.

This issue may exist for other configuration properties as well, I've been testing and investigating using the step size.

Chris927 (changes requested):

resources/victron-virtual-functions.js
dirkjanfaber: You are right. I noticed that too and have it changed in the next commit (https://github.com/victronenergy/node-red-contrib-victron/pull/304/commits/d9f7729429a3e49c9b5a12d8c821053673ac9587#diff-3d66eb160ddb4af40902a58e0b968a9cf99469cf3d984cbeff7d4b05e2341383L93). 

And indeed, left the 1 there for backwards compatibility. In the post on community we already advice people to go back from multiple switches to just one. So this will save at least those who do that from doing another change.

src/nodes/victron-virtual.js
dirkjanfaber: Yes, that looks like it will solve it.

src/nodes/victron-virtual.js
dirkjanfaber: Probably the best way to solve this is to add a disclaimer on the edit panel. E.g. with something like:
> Note that if you change these settings via a control node, the configuration value shown here might be out of sync.

As this is only relevant for users that actually do change these values, they are probably also fine with accepting that the reasoning for the values getting out of sync.

dirkjanfaber: Resolving this by mentioning it in issue #240, to make sure it ends up in the documentation.

PR #310, opened 2025-10-12
fix: properly release DBus service name on node close

dirkjanfaber:
When a virtual device node is deleted during re-deploy, the DBus service name was not being released, causing the service to remain registered on the bus until Node-RED fully restarts.

Changes:
- Store serviceName on node instance after successful requestName
- Call bus.invoke() to invoke DBus ReleaseName before closing connection
- Ensure ReleaseName completes before calling connection.end()

Note: this might be better solved in the underlying library instead, exposing a call for the ReleaseName.

### DIFF
--- a/src/nodes/victron-virtual.js
+++ b/src/nodes/victron-virtual.js
@@ -352,6 +352,7 @@ module.exports = function (RED) {
       this.address = `tcp:host=${address[0]},port=${address[1]}`
     }
 
+    node.retryOnConnectionEnd = true
     node.pendingCallsToSetValuesLocally = []
 
     function handleInput (msg, done) {
@@ -464,7 +465,11 @@ module.exports = function (RED) {
           text: 'DBus connection closed'
         })
         debugConnection('DBus connection closed, retrying...')
-        retryConnectionDelayed()
+        if (self.retryOnConnectionEnd) {
+          retryConnectionDelayed()
+        } else {
+          debug('Not retrying DBus connection, as retryOnConnectionEnd is false.', interfaceName)
+        }
       })
 
       self.bus.connection.on('error', (err) => {
@@ -1321,6 +1326,7 @@ module.exports = function (RED) {
         // were left. Calling end() here resolves an issue with the VictronDbusListener
         // not responding to ItemsChanged signals any more after a redeploy here:
         // https://github.com/victronenergy/node-red-contrib-victron/blob/5626b44b426a3ab1c7d9a6a2d36f035f72d9faa2/src/services/dbus-listener.js#L309
+        node.retryOnConnectionEnd = false
         node.bus.connection.end()
 
         // If this was the last instance and the timeout is still pending


src/nodes/victron-virtual.js
```diff
@@ -1292,22 +1294,56 @@ module.exports = function (RED) {
...
+          path: '/org/freedesktop/DBus',
+          destination: 'org.freedesktop.DBus',
+          interface: 'org.freedesktop.DBus',
+          member: 'ReleaseName',
+          signature: 's',
+          body: [node.serviceName],
+          type: 1 // dbus.messageType.methodCall
+        }, (err, result) => {
+          if (err) {
+            debug(`Error releasing service name ${node.serviceName}:`, err)
```
Chris927: Rather `console.error` instead of `debug`, so that errors when releasing the name won't be silent.

src/nodes/victron-virtual.js
```diff
@@ -1292,22 +1294,56 @@ module.exports = function (RED) {
...
+          } else {
+            const releaseCode = result?.[0]
```
Chris927: When testing, `result?.[0]` evaluated to `undefined`, `result` was `1`. Why are you using `result?.[0]`, it seems `result` is correct.

src/nodes/victron-virtual.js
```diff
@@ -1292,22 +1294,56 @@ module.exports = function (RED) {
...
-          clearTimeout(globalTimeoutHandle)
-          globalTimeoutHandle = null
-        }
-        hasRunOnce = false
+      // Release the DBus service name before closing connection
+      // This prevents the service from remaining registered after node deletion
+      if (node.serviceName && this.bus && this.bus.invoke) {
+        debug(`Releasing DBus service name: ${node.serviceName}`)
+
+        this.bus.invoke({
```
Chris927: Why not use `this.bus.releaseName()`, then it would be analogous to calling `this.bus.requestName()`.

As I understand your comment in the PR description, this is your idea already :) as per [this](https://github.com/Chris927/dbus-native/blob/0b04da3f37d30b12d45bc8ccc1a5687c94355892/lib/bus.js#L351), the function `this.bus.releaseName()` should be available.

src/nodes/victron-virtual.js
dirkjanfaber: Resolved in [this](https://github.com/victronenergy/node-red-contrib-victron/pull/310/commits/fee5c4557c56104eab0faf54cf9c371ed0800e31) commit.

src/nodes/victron-virtual.js
dirkjanfaber: Yes, changed that in [this](https://github.com/victronenergy/node-red-contrib-victron/pull/310/commits/fee5c4557c56104eab0faf54cf9c371ed0800e31) commit. 
While the PR seem to work, directly after the service gets re-created on the dbus still. Not sure what causes that.

PR #311, opened 2025-10-12
Refactor code, part 1

dirkjanfaber:
This PR ended up a bit bigger than I originally anticipated. So there is quite a lot going on. 

In short it adds documentation and (dynamic) outputs to the virtual devices, with the main focus on the switches. The non-switches probably need extra outputs as well, but that will be for a new branch. 

Also pulled out some code for better maintainability.

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -376,8 +376,6 @@
             <input type="${field.type}" id="node-input-switch_1_${field.id}"
                    placeholder="${field.placeholder}"
                    style="${field.style}"
-                   ${field.min ? `min="${field.min}"` : ''}
-                   ${field.max ? `max="${field.max}"` : ''}
                    ${stepAttr} required>
           </div>
         `
@@ -397,6 +395,28 @@
           const val = context[`switch_1_${field.id}`];
           if (typeof val !== 'undefined') {
             $(`#node-input-switch_1_${field.id}`).val(val);
+          }
+
+          const $input = $(`#node-input-switch_1_${field.id}`);
+
+          // Add validation for stepped switch max field
+          if (field.id === 'max' && Number(type) === victronVirtualConstantsExports.SWITCH_TYPE_MAP.STEPPED) {
+            $input.on('blur input', function () {
+              const val = $(this).val();
+              const maxVal = parseInt(val, 10);
+
+              if (!val) {
+                // Empty field - will be caught by required validation
+                this.setCustomValidity('This field is required');
+                this.reportValidity();
+              } else if (isNaN(maxVal) || maxVal < 1 || maxVal > 7) {
+                this.setCustomValidity('Max steps must be between 1 and 7');
+                this.reportValidity();
+              } else {
+                // Valid - clear any error
+                this.setCustomValidity('');
+              }
+            });
           }
         });
 
@@ -451,7 +471,7 @@
         }
       }
 
-      if (Number(type) === victronVirtualConstantsExports.SWITCH_TYPE_MAP.TEMP_SETPOINT) {
+      if (Number(type) === victronVirtualConstantsExports.SWITCH_TYPE_MAP.TEMPERATURE_SETPOINT) {
         // Add checkbox for Measurement path
         const measurementToggle = $(`
         <div class="form-row" id="switch-1-measurement-toggle-row">
@@ -514,11 +534,17 @@
     }
   }
 
-  function updateSwitchConfig () {
+  function updateSwitchConfig (context) {
     const container = $('#switch-config-container');
     container.empty();
     if ($('select#node-input-device').val() !== 'switch') return
-    renderSwitchConfigRow(this);
+    renderSwitchConfigRow(context);
+
+    // Add handler for switch type changes
+    $('#node-input-switch_1_type').on('change', (v) => {
+      context.switch_1_type = v.target.value;
+      updateOutputs(context);
+    });
   }
 
   function updateBatteryVoltageVisibility () {
@@ -533,7 +559,7 @@
     $('#battery-voltage-custom-label').toggle(preset === 'custom');
   }
 
-  function checkSelectedVirtualDevice () {
+  function checkSelectedVirtualDevice (context) {
     [
       'battery', 'generator', 'gps', 'grid', 'motordrive', 'pvinverter',
       'switch', 'tank', 'temperature'
@@ -574,7 +600,7 @@
     }
 
     if (selected === 'switch') {
-      updateSwitchConfig.call(this);
+      updateSwitchConfig(context);
       // Hide the default values checkbox and info box for switches
       $('#node-input-default_values').closest('.form-row').hide();
       $('#default-values-info').hide();
@@ -658,11 +684,10 @@
    * @param {object} context - Node-RED editor context (this)
    */
   function updateOutputs (context) {
-    const device = $('#node-input-device').val();
+    const device = context.device;
     const config = {
-      switch_1_type: $('#node-input-switch_1_type').val()
+      switch_1_type: context.switch_1_type
     };
-
     const outputs = calculateOutputs(device, config);
 
     // Update BOTH the context AND the hidden input field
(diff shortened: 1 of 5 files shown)


src/nodes/victron-virtual.js
```diff
@@ -825,14 +848,14 @@ module.exports = function (RED) {
               }
               if (name === 'Settings/Type') propValue = switchType
 
-              iface[switchOutputPropertyKey] = propValue !== undefined ? propValue : 0
+              iface[switchableOutputPropertyKey] = propValue !== undefined ? propValue : 0
 
               if (name === 'Settings/ValidTypes') {
                 // Only allow the currently selected switch type in the GUI.
                 // This sets /ValidTypes to a bitmask with only the current type allowed.
                 // See: https://github.com/victronenergy/dbus-victron-virtual/issues/XXX
```
Chris927: `XXX` does not make sense.

resources/victron-virtual-functions.js
```diff
@@ -74,61 +126,268 @@
...
       label: 'Bilge pump control',
       fields: [...COMMON_SWITCH_FIELDS]
     }
   };
 
+  const SWITCH_TYPE_DOCS = {
+    [victronVirtualConstantsExports.SWITCH_TYPE_MAP.MOMENTARY]: {
+      text: `
+      <div>
+        <strong>Most relevant path(s):</strong>
```
Chris927, line 143: Why have a section `Most relevant path(s)`? Analogous to `Outputs`, we could list `Inputs`.

Related: Is there a place where I can see which properties are writeable, and which ones are read-only? When testing, I'm just guessing what might be writable.

It may also help to document here how to set the inputs: The simplest I can come up with is a function node with:

```javascript
msg.payload = {
    'SwitchableOutput/output_1/State': 1
};
return msg;
```

Maybe there is an easier way?

resources/victron-virtual-functions.js
```diff
@@ -142,7 +401,7 @@
         });
 
         // Special handling for dropdown type
-        if (type === '6') {
+        if (String(type) === String(victronVirtualConstantsExports.SWITCH_TYPE_MAP.DROPDOWN)) {
```
Chris927, line 424: It's awesome to not have magic numbers (like `6` here) any more!

resources/victron-virtual-functions.js
```diff
@@ -281,6 +583,13 @@
 
     if (selected === 'switch') {
       updateSwitchConfig.call(this);
```
Chris927: Why do you use `call()` here? Why not use the same pattern as for e.g. `renderSwitchConfigRow(context)`? Then we don't rely on the (IMHO) odd semantics of `.call()` :)

resources/victron-virtual-functions.js
Chris927: Further, to keep it DRY, we could have a text constant, and then refer to it in the documentation texts:

```javascript

const HOW_TO_WRITE_SWITCH_STATE = `
<strong>How to write the Switch State</strong>
<p>In order to set the switch state, you can use a function that sets the payload, like this:</p>
<pre>
msg.payload = {
    'SwitchableOutput/output_1/State': 1
};
return msg;
</pre>`

...
    text: ` ... ${HOW_TO_WRITE_SWITCH_STATE} ...`
```

resources/victron-virtual-functions.js
```diff
@@ -173,6 +432,40 @@
...
+      $('#switch-1-doc-row').remove();
+      if (doc) {
+        const docRow = $(`
+        <div id="switch-1-doc-row" class="victron-doc-box">
+          <label>${cfg.label} usage</label>
+          ${doc.img ? `<img src="${doc.img}" alt="Switch type preview">` : ''}
+          <div class="victron-doc-text">${doc.text}</div>
+        </div>
+      `);
+        // Append after the options row if it exists, otherwise after the config row
```
Chris927, line 466: [not essential to change] An alternative to requiring if/else logic here: You could always have an element with the ID `switch-1-pairs-row`, but in case it is not required, its content could be empty.

resources/victron-virtual-functions.js
```diff
@@ -225,6 +519,14 @@
     container.empty();
     if ($('select#node-input-device').val() !== 'switch') return
     renderSwitchConfigRow(this);
+
+    console.log("break here");
```
Chris927: remove this

resources/victron-virtual-functions.js
```diff
@@ -296,8 +605,9 @@
           $input[0].reportValidity();
           return false
         } else if ($input.length) {
-          if (field.id === 'max' && type === '4') {
-            const maxVal = parseInt($input.val(), 10);
+          if (field.id === 'max' && Number(type) === victronVirtualConstantsExports.SWITCH_TYPE_MAP.STEPPED) {
+            const val = $input.val();
+            const maxVal = parseInt(val, 10);
             if (isNaN(maxVal) || maxVal < 1 || maxVal > 7) {
               $input[0].setCustomValidity('Max steps must be between 1 and 7');
```
Chris927, line 630: When I set the value to 8 in the config form, the validation text does not show up anywhere. The border of the input HTML element turns red, though.

src/nodes/victron-virtual-browser.js
```diff
@@ -1,17 +1,26 @@
...
 window.SWITCH_TYPE_CONFIGS = SWITCH_TYPE_CONFIGS
+window.SWITCH_OUTPUT_CONFIG = SWITCH_OUTPUT_CONFIG
+window.SWITCH_THIRD_OUTPUT_LABEL = SWITCH_THIRD_OUTPUT_LABEL
 window.renderSwitchConfigRow = renderSwitchConfigRow
 window.updateSwitchConfig = updateSwitchConfig
 window.checkSelectedVirtualDevice = checkSelectedVirtualDevice
 window.validateSwitchConfig = validateSwitchConfig
 window.updateBatteryVoltageVisibility = updateBatteryVoltageVisibility
+window.calculateOutputs = calculateOutputs
+window.updateOutputs = updateOutputs
```
Chris927, line 26: We are adding a lot of functions to `window`, a global object. If another package does something similar (e.g. also add a function `updateOutputs` to `window`, then we get weird behavior which will be hard to trace.

An option is to put our global functions into `window.__victron`, e.g. `window.__victron.updateOutputs`, which will however require extra work to access those functions (as they won't be immediately available as globals).

Maybe don't change this now, but add an issue for us to clean this up later.

resources/victron-virtual-functions.js
dirkjanfaber: If you hover above it, it shows a different message than this. 😕 

image

resources/victron-virtual-functions.js
dirkjanfaber: Totally agree!

resources/victron-virtual-functions.js
dirkjanfaber: Resolved that in [this](fb4a7db4c83e0f987f66896744bf135d90f438e1) commit. 

image

resources/victron-virtual-functions.js
dirkjanfaber: Yes, I agree. Looking in how to properly get rid if this. Feel free to suggest the best way to do this.

resources/victron-virtual-functions.js
dirkjanfaber: I opted to have most relevant paths as some of the rest of the paths (as described on https://github.com/victronenergy/venus/wiki/dbus#switch) are not that useful to modify during runtime. 

But we could refer to the full documentation. I suggest awaiting user feedback on this.

src/nodes/victron-virtual.js
dirkjanfaber: There was a gui-v2 issue for this where I wanted to refer to, but couldn't find it when I wrote that comment. Still can't find it, so leaving it out.

dirkjanfaber: See 435796dcc31e14270114357743e16875cbec42aa

src/nodes/victron-virtual-browser.js
dirkjanfaber: To be resolved via #315

resources/victron-virtual-functions.js
dirkjanfaber: Resolved via this commit: a98c04f36685b4261115d764062d292abb494374

resources/victron-virtual-functions.js
dirkjanfaber: Yes, perhaps at some later point. I think this is good for now.

PR #317, opened 2025-10-17
Update documentation and input validation

dirkjanfaber:
User Andre on community failed to notice that the virtual device was exxpecting a JSON object. This change helps clarify that in both input checking as validation.

See https://community.victronenergy.com/t/venus-os-v3-70-49-and-virtual-switch-in-node-red/47227 for more details and some screenshots.

Also added feedback of luphi in the second commit.

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -33,14 +33,13 @@
   	  [SWITCH_TYPE_MAP.BILGE_PUMP]: 2 // passthrough + state
   	};
 
+  	// Will default to 'State' if not defined here
   	const SWITCH_SECOND_OUTPUT_LABEL = {
   	  [SWITCH_TYPE_MAP.TEMPERATURE_SETPOINT]: 'Temperature',
   	  [SWITCH_TYPE_MAP.DROPDOWN]: 'Selected',
   	  [SWITCH_TYPE_MAP.BASIC_SLIDER]: 'Value'
-  	  // All other switch types use 'State' as default
   	};
 
-  	// Output labels configuration: third label varies by switch type
   	const SWITCH_THIRD_OUTPUT_LABEL = {
   	  [SWITCH_TYPE_MAP.DIMMABLE]: 'Dimming',
   	  [SWITCH_TYPE_MAP.STEPPED]: 'Value',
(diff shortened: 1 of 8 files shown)


src/nodes/victron-virtual.js
```diff
@@ -368,24 +368,55 @@ module.exports = function (RED) {
...
-          text: 'Invalid payload - passthrough only'
+          text: `Invalid payload (${receivedType}) - expected JSON object`
+        })
+        done()
+        return
+      }
+
+      // Check if object is empty
+      if (Object.keys(msg.payload).length === 0) {
+        node.warn('Received empty object. Expected: JSON object with {path: value} pairs.')
```
Chris927: `JSON object with {path: value} pairs.` is slightly odd wording: There is no JSON involved (at this point), it's a plain JavaScript object. Maybe change to `JavaScript object with at least one property/value`.

The same wording issue is present in some other places.

resources/victron-virtual-functions.js
```diff
@@ -136,210 +142,78 @@
...
-        <strong>Outputs:</strong>
-        <ol>
-          <li><code>Passthrough</code> &mdash; Outputs the original <tt>msg.payload</tt> without modification</li>
-          <li><code>State</code> &mdash; <tt>msg.payload</tt> contains a <tt>0</tt> or <tt>1</tt> representing the on/off state of the pump</li>
-        </ol>
-      </div>
-    `,
-      img: '/resources/@victronenergy/node-red-contrib-victron/docs/bilge_pump.png'
-    }
+    [victronVirtualConstantsExports.SWITCH_TYPE_MAP.MOMENTARY]: createDocTemplate(
```
Chris927, line 164: The object `victronVirtualConstantsExports.SWITCH_TYPE_MAP` and its properties are referenced a lot. I suggest to make this more readable by assigning the values of the `SWITCH_TYPE_MAP` once at the top, then use those variables:

```javascript
const { MOMENTARY, TOGGLE, ... } = victronVirtualConstantsExports.SWITCH_TYPE_MAP;
```

This simplifies this line to:

```javascript
[MOMENTARY]: createDocTemplate(
```

Which is easier to read.

src/nodes/victron-virtual.html
```diff
@@ -73,28 +73,43 @@
...
             }
-            
+
             const switchType = this.switch_1_type
             const typeKey = switchType !== undefined ? parseInt(switchType, 10) : 1 // TOGGLE = 1
             
-            // Map of third output labels
+            // Map of second output labels (for switches that don't have "State")
+            const secondLabels = {
+                3: 'Temperature',  // TEMPERATURE_SETPOINT
```
Chris927, line 82: Can you avoid the magic number `3`? I guess it could be:

```javascript
const secondLabels = {
  [TEMPERATURE_SETPOINT]: 'Temperature',
...
```

The same applies to a few other places in this PR.

Chris927 (dismissed):

resources/victron-virtual-functions.js
dirkjanfaber: True, but most of that is generated via the `npm build` step, creating the `resources` file.

src/nodes/victron-virtual.html
dirkjanfaber: That is in the outputLabels section, which has `// Inline logic since window.getOutputLabels may not be loaded yet`, so it is the fallback for when the numbers aren't known yet.

src/nodes/victron-virtual.js
dirkjanfaber: Resolved via [497a00d](https://github.com/victronenergy/node-red-contrib-victron/pull/317/commits/497a00dc9736ac3a53181d734d5d1bc5de9fbf4e)

PR #324, opened 2025-11-05
fix: prevent global namespace pollution with window.__victron

### LINKED ISSUE #315: Put global functions in more private namespace
**Describe the bug**
We are adding a lot of functions to window, a global object. If another package does something similar (e.g. also add a function updateOutputs to window, then we get weird behavior which will be hard to trace.

An option is to put our global functions into window.__victron, e.g. window.__victron.updateOutputs, which will however require extra work to access those functions (as they won't be immediately available as globals).

(As remarked in [PR #311](https://github.com/victronenergy/node-red-contrib-victron/pull/311#discussion_r2432647450) )

dirkjanfaber:
Closes #315

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -617,9 +617,6 @@
   window.__victron = {
     checkGeneratorType,
     SWITCH_TYPE_CONFIGS,
-    SWITCH_OUTPUT_CONFIG: victronVirtualConstantsExports.SWITCH_OUTPUT_CONFIG,
-    SWITCH_SECOND_OUTPUT_LABEL: victronVirtualConstantsExports.SWITCH_SECOND_OUTPUT_LABEL,
-    SWITCH_THIRD_OUTPUT_LABEL: victronVirtualConstantsExports.SWITCH_THIRD_OUTPUT_LABEL,
     renderSwitchConfigRow,
     updateSwitchConfig,
     checkSelectedVirtualDevice,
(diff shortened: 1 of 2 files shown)


resources/victron-virtual-functions.js
```diff
@@ -614,17 +614,19 @@
...
-  window.checkSelectedVirtualDevice = checkSelectedVirtualDevice;
-  window.validateSwitchConfig = validateSwitchConfig;
-  window.updateBatteryVoltageVisibility = updateBatteryVoltageVisibility;
-  window.calculateOutputs = calculateOutputs;
-  window.updateOutputs = updateOutputs;
+  window.__victron = {
+    checkGeneratorType,
+    SWITCH_TYPE_CONFIGS,
+    SWITCH_OUTPUT_CONFIG: victronVirtualConstantsExports.SWITCH_OUTPUT_CONFIG,
+    SWITCH_SECOND_OUTPUT_LABEL: victronVirtualConstantsExports.SWITCH_SECOND_OUTPUT_LABEL,
```
Chris927: Is `SWITCH_SECOND_OUTPUT_LABEL` required in `window.__victron`? From my understanding it isn't.

I removed it from `window.__victron`, ran unit tests, and tested manually, all seems fine.

dirkjanfaber: That is a bit outside of the scope of this PR, but you are right. That doesn't seem to be needed.  The same would go for `SWITCH_OUTPUT_CONFIG` and `SWITCH_THIRD_OUTPUT_LABEL` then. 

I'll update the branch to remove those in a separate commit.

Chris927 (approved):

PR #325, opened 2025-11-05
feat: add rate limiting for input nodes

### LINKED ISSUE #314: input-gridmeter update frequency
**Describe the bug**
This probably should be fixed on the updating on the dbus side, but perhaps there is a way to limit the nr of updates here too.

See https://community.victronenergy.com/t/venus-os-v3-70-45-available-for-public-testing/45708/101

This leads into a Memory Leak warning.

**Steps to reproduce**

Use a VM 375 and read (and plot) the high frequency updates.

dirkjanfaber:
Adds configurable rate limiting to all input nodes to handle high-frequency updates from devices like grid meters. Users can select from preset rates (10/5/2/1 msg/sec, or 1 per 2/5 sec) via dropdown. Defaults to no limit.

Resolves #314

### DIFF
--- a/src/nodes/victron-nodes.js
+++ b/src/nodes/victron-nodes.js
@@ -53,8 +53,6 @@ module.exports = function (RED) {
       this.onlyChanges = nodeDefinition.onlyChanges
       this.roundValues = nodeDefinition.roundValues
       this.rateLimit = nodeDefinition.rateLimit || 0
-      this.throttleMs = this.rateLimit > 0 ? Math.floor(1000 / this.rateLimit) : 0
-      this.lastSentTime = 0
       this.sentInitialValue = false
 
       this.configNode = RED.nodes.getNode('victron-client-id')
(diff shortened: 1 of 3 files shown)


src/nodes/victron-nodes.js
```diff
@@ -79,6 +82,11 @@ module.exports = function (RED) {
         const isPollingEnabled = this.configNode.enablePolling || false
         const callbackPeriodically = !this.node.onlyChanges && !isPollingEnabled
         this.subscription = this.client.subscribe(this.service, this.path, (msg) => {
+          const now = Date.now()
+          if (this.node.throttleMs > 0 && this.sentInitialValue && (now - this.node.lastSentTime) < this.node.throttleMs) {
+            return
```
Chris927: This will result in the latest msg being ignored, when throttling. Example, assuming throttling at 1 message / second:

1. at 0ms, we get value 1, not throttled.
2. at 1ms, we get value 2, throttled.
3. No more messages.

Because we `return` here, we will never process value 2.

What we rather need is throttling the *calls* without throttling the *values*.

This can be achieved with a higher-order function, see [this feature branch](https://github.com/Chris927/node-red-contrib-victron/commit/db5823383567583693bc795f5b8fdf7b56572a3d); the feature branch should be mergeable into `issue-314`.

dirkjanfaber: Thanks for the feature branch. I've taken that and updated mine.

PR #328, opened 2025-11-06
feat: support ShowUIControl bitmask values 0-6

dirkjanfaber:
See https://github.com/victronenergy/venus/issues/1554:

The `/ShowUiContro`l path has been updated to an integer value:

Path values:
- `0b000`: Do not show the control item
- `0bxx1:` Show in all UI's
- `0bx10:` Show in the local UI
- `0b1x0`: Show in all remote UI's

The default behavior should be to show it in all UI's, so the value should be set to 1. This is currently also the default behavior. Only thing that needs to be checked is whether values other than 0 and 1 are accepted by the driver. A value of up to 0b110 = 6 must be accepted.

### DIFF
--- a/test/victron-virtual-showuicontrol.test.js
+++ b/test/victron-virtual-showuicontrol.test.js
@@ -1,0 +1,63 @@
+// test/victron-virtual-showuicontrol.test.js
+// Tests for ShowUIControl bitmask support
+
+describe('ShowUIControl bitmask support', () => {
+  test('should accept valid bitmask values 0-6', () => {
+    const validValues = [0, 1, 2, 3, 4, 5, 6]
+
+    validValues.forEach(value => {
+      expect(value).toBeGreaterThanOrEqual(0)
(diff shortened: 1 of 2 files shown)


test/victron-virtual-showuicontrol.test.js
```diff
@@ -0,0 +1,63 @@
+// test/victron-virtual-showuicontrol.test.js
+// Tests for ShowUIControl bitmask support
+
+describe('ShowUIControl bitmask support', () => {
```
Chris927, line 4: These tests are not testing any "unit", they are testing that JavaScript language features work as expected.

It would be more useful to externalize the `format` function of `Settings/ShowUIControl`, and verify via unit tests that for all inputs to the `format()` function the `format()` function returns the expected output.

Chris927 (approved):
We should adjust the unit test, as per my comment. The change itself should be mergeable now regardless, that's why I'm approving.

PR #329, opened 2025-11-06
feat: debounce virtual device property writes to "fix" UI race conditions

dirkjanfaber:
Gui-v2 sometimes set the Type to `-1` (not sure yet why it does that). But this can cause a race condition where the type quickly switches from `-1` to the value the type should have, e.g. `11`. This seems to be triggered sometimes at (re)deploy of the flow and gives a bad user experience.

Even though this should be solved at the root, reducing the speed where the drive accepts changes of a value could already help. So adding a default 300 ms delay after a change for a subsequent one should help. This is by default enabled for _all_ paths, but can be overruled by setting the `immediate: true` option.

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -46,11 +46,15 @@
   	  [SWITCH_TYPE_MAP.NUMERIC_INPUT]: 'Value'
   	};
 
+  	// Default debounce delay for virtual device property writes (in milliseconds)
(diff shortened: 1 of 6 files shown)


Chris927 (changes requested):
See comment [here](https://github.com/victronenergy/node-red-contrib-victron/pull/329#issuecomment-3508797290).

PR #341, opened 2025-11-26
feat: add tooltips and community link to input/output nodes

dirkjanfaber:
- Extract tooltip system from virtual device switches to shared utility
- Add info icon tooltips to form fields
- Add community link button at bottom of node edit dialogs
- Improve form field alignment
- Enhance "Currently stored selection" display with info icon
- Add communityTag to services.json for generating the community links (not used yet)
- Use .svg instead of .png files

This is all part of unifying the documentation and making it more consistent across the nodes.

### DIFF
--- a/resources/victron-virtual-style.css
+++ b/resources/victron-virtual-style.css
@@ -15,7 +15,7 @@ a.blue-link {
   vertical-align: middle;
 }
 
-.form-row input[type="checkbox"] {
+.victron-form .form-row input[type="checkbox"] {
   width: auto !important;
 }
 


src/nodes/victron-nodes.html
```diff
@@ -310,7 +317,9 @@
                 }
             }
         })
-        .done(data => {
+        .done(response => {
+            communityTag = response.communityTag;
+            var data = response.services || [];
```
Chris927, line 327: This breaks the display of custom input services: There is no property `services` in the response, it's just an array of services. As a result, the list of available services is always empty.

So probably `response || []` is correct. This will likely break tests, though.

(This relates to our Slack discussion, @dirkjanfaber .)

Chris927 (changes requested):

src/nodes/victron-nodes.html
dirkjanfaber: Thanks for finding this. I resolved it slightly different (see https://github.com/victronenergy/node-red-contrib-victron/pull/341/commits/ec80a8f663f21096c3d3db7e8feaf2b9e6b1c2f0).

resources/victron-virtual-style.css
```diff
@@ -1,48 +1,68 @@
...
+
+.form-row input[type="text"],
+.form-row input[type="number"],
+.form-row select {
+  width: 240px !important;
+  vertical-align: middle;
+}
+
+.form-row input[type="checkbox"] {
+  width: auto !important;
```
Chris927, line 19: This breaks the "gap" between the checkbox input and the label, i.e. the little gap is now gone.

This is before:

image

... and this is after:

image

This affects at least the debug node. For virtual device nodes it seems to be okay (well, kind-of, the checkbox is a little vertically offset...):

image

I don't understand the structure of css styling well enough to suggest a solution 😅

Chris927 (changes requested):

resources/victron-virtual-style.css
dirkjanfaber: I have resolved that issue. Mainly by putting the victron edit panel in its own css scope.

src/nodes/victron-nodes.html
```diff
@@ -175,25 +166,44 @@
...
+            }
             updateEnums();
+            updateCommunityLink();
+        };
+
+        var communityTag = null;  // Store communityTag at module level
+
+        var updateCommunityLink = function updateCommunityLink() {
+            if (communityTag) {
+                var url = 'https://community.victronenergy.com/c/node-red/28/'; // + communityTag;
```
Chris927, line 185: Is the `communityTag` commented out on purpose? I noticed that for `battery` there is no such tag on the Forum, but there are multiple tags related to batteries (`battery_balancer`, `battery_charging`, etc).

If this is not ready yet, maybe add a comment indicating this.

resources/victron-virtual-style.css
```diff
@@ -1,48 +1,68 @@
...
+}
+
+.form-row input[type="text"],
+.form-row input[type="number"],
+.form-row select {
+  width: 240px !important;
+  vertical-align: middle;
+}
+
+.form-row input[type="checkbox"] {
```
Chris927: As I understand the design guide, these rules should all be pre-fixed with `.victron-form`, no?

Further, I noticed that the checkboxes on the debug node are still without the space between checkbox and label (probably because of this):

Image

dirkjanfaber: Resolved that by putting the fix you suggested. But that isn't really solving the problem. There are still 2 css files and not every entry is scoped under `.victron-form` (yet). Lets solve that via a new issue as a couple of hours of trying to solve all of these styling issues didn't work yet.

dirkjanfaber: Created issue #343 for that.

src/nodes/victron-nodes.html
dirkjanfaber: Added comments to clarify.

Issue #335, opened 2025-11-13
[Feature] inject notification node


dirkjanfaber:
With the new notification center, we are able to inject notifications.

Inject  `msg.payload` should be something like:
```
{
  "type": 0,
  "title": "Hello world",
  "message": "Node-RED on Victron Energy devices is cool!"
}
```

Which then translates to a `SetValue` `"${type}\t${title}\t${message}"` on `com.victronenergy.platform` `/Notifications/Inject`

So this will need its own control node.

dirkjanfaber:
Closing as this has been implemented (release 1.6.56)

PR #344, opened 2025-11-30
Add inject node (issue #335)

dirkjanfaber:
This is a new node for injecting notification messages from Node-RED. 

**Usage:**
Send the notification message in `msg.payload`.

Optionally override the configured type with `msg.type` (number: 0/1/2 or string: "warning"/"alarm"/"info").

Optionally override the configured title with `msg.title`.

The output includes `msg.notification` with injected notification details.

Status displays "Title: message" with color-coded dot (yellow=Warning, red=Alarm, blue=Information).

### DIFF
--- a/src/nodes/victron-inject-functions.js
+++ b/src/nodes/victron-inject-functions.js
@@ -45,20 +45,10 @@ function validateNotificationInput (payload, type, title) {
 
   const MAX_TITLE_LENGTH = 100
   const MAX_MESSAGE_LENGTH = 500
+  const TYPE_ERROR = 'Type must be "warning", "alarm", "info"/"information", or 0-2'
 
-  if (String(title).length > MAX_TITLE_LENGTH) {
-    return {
-      valid: false,
-      error: `Title too long (max ${MAX_TITLE_LENGTH} chars)`
-    }
-  }
-
-  if (message.length > MAX_MESSAGE_LENGTH) {
-    return {
-      valid: false,
-      error: `Message too long (max ${MAX_MESSAGE_LENGTH} chars)`
-    }
-  }
+  const truncatedTitle = String(title).substring(0, MAX_TITLE_LENGTH)
+  const truncatedMessage = message.substring(0, MAX_MESSAGE_LENGTH)
 
   let typeNum
   if (typeof type === 'string') {
@@ -74,29 +64,29 @@ function validateNotificationInput (payload, type, title) {
     } else {
       return {
         valid: false,
-        error: 'Type must be "warning", "alarm", "info"/"information", or 0-2'
+        error: TYPE_ERROR
       }
     }
   } else if (typeof type === 'number') {
     if (!Number.isInteger(type) || type < 0 || type > 2) {
       return {
         valid: false,
-        error: 'Type must be "warning", "alarm", "info"/"information", or 0-2'
+        error: TYPE_ERROR
       }
     }
     typeNum = type
   } else {
     return {
       valid: false,
-      error: 'Type must be "warning", "alarm", "info"/"information", or 0-2'
+      error: TYPE_ERROR
     }
   }
 
   return {
     valid: true,
     type: typeNum,
-    title: String(title),
-    message
+    title: truncatedTitle,
+    message: truncatedMessage
   }
 }
 
--- a/src/nodes/victron-inject.html
+++ b/src/nodes/victron-inject.html
@@ -62,6 +62,7 @@
                 <p>Optionally override the configured type with <code>msg.type</code> (number: 0/1/2 or string: "warning"/"alarm"/"info").</p>
                 <p>Optionally override the configured title with <code>msg.title</code>.</p>
                 <p>The output includes <code>msg.notification</code> with injected notification details.</p>
+                <p>Title and message are automatically truncated to 100 and 500 characters respectively.</p>
                 <p>Status displays "Title: message" with color-coded dot (yellow=Warning, red=Alarm, blue=Information).</p>
             </div>
         </div>
@@ -126,9 +127,8 @@ The node will inject the notification into the dbus at `com.victronenergy.platfo
 
 The node passes through the original message and adds a `msg.notification` object containing:
 - `type`: The type number used (0, 1, or 2)
-- `title`: The title string used
-- `message`: The message string used
-- `timestamp`: ISO timestamp of when the notification was injected
+- `title`: The title string used (truncated to 100 characters if longer)
+- `message`: The message string used (truncated to 500 characters if longer)
 
 ```javascript
 {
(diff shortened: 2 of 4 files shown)


examples/Inject Notifications.json
```diff
@@ -0,0 +1,168 @@
...
+                "notification-basic"
+            ]
+        ]
+    },
+    {
+        "id": "notification-basic",
+        "type": "victron-inject",
+        "z": "example-flow-id",
+        "name": "System Info",
+        "notificationType": "2",
```
Chris927, line 42: When testing this flow, I get `Error: Type must be "warning", "alarm", "info"/"information", or 0-2`, because `validateNotificationInput()` sees a string (and then there is no match in `typeMap`).

When I remove the notification node and re-create it, I don't get this error, and the value for `notificationType` in the exported JSON is `2`, not `"2"`.

Maybe this json file was exported before a code change?

I guess changing this from `"2"` to `2` should fix this.

examples/Inject Notifications.json
```diff
@@ -0,0 +1,168 @@
...
+                "check-soc"
+            ]
+        ]
+    },
+    {
+        "id": "check-soc",
+        "type": "function",
+        "z": "example-flow-id",
+        "name": "Check SOC threshold",
+        "func": "const soc = msg.payload;\n\nif (soc < 20) {\n    msg.payload = `Battery state of charge is critically low: ${soc}%`;\n    msg.type = \"alarm\"; // Alarm (can also use 1)\n    msg.title = \"Battery Critical\";\n    return msg;\n} else if (soc < 30) {\n    msg.payload = `Battery state of charge is low: ${soc}%`;\n    msg.type = \"warning\"; // Warning (can also use 0)\n    msg.title = \"Battery Low\";\n    return msg;\n}\n\n// Don't send notification if SOC is OK\nreturn null;",
```
Chris927, line 121: Is there a way to clear a notification? The issue with a use case like this is that the alarm / warning would persist, even though the reason for the alarm is no more.

If there is no way yet to clear, maybe there could be a note about it in the docs to not raise expectations, and we should have an issue on the backlog to clear notifications (somehow).

src/nodes/victron-inject-functions.js
```diff
@@ -0,0 +1,106 @@
...
+      valid: false,
+      error: 'Notification title is required'
+    }
+  }
+
+  const MAX_TITLE_LENGTH = 100
+  const MAX_MESSAGE_LENGTH = 500
+
+  if (String(title).length > MAX_TITLE_LENGTH) {
+    return {
```
Chris927: This is a bad failure mode: Imagine my title is something like "warning ... 99 entries", and the length is exactly `MAX_TITLE_LENGTH`, then I get one more entry (whatever an entry is), the title becomes "warning ... 100 entries", and suddenly the notification fails, because I am now past the max, resulting in no notification being shown.

I suggest to trim title and message, if necessary, but not make the notification fail because of any length being exceeded.

src/nodes/victron-inject-functions.js
```diff
@@ -0,0 +1,106 @@
...
+      information: 2,
+      info: 2
+    }
+    const typeLower = type.toLowerCase()
+    if (typeLower in typeMap) {
+      typeNum = typeMap[typeLower]
+    } else {
+      return {
+        valid: false,
+        error: 'Type must be "warning", "alarm", "info"/"information", or 0-2'
```
Chris927: The same text is repeated, maybe move to a local const `TYPE_ERROR_TEXT` or similar.

src/nodes/victron-inject.html
```diff
@@ -0,0 +1,153 @@
...
+Send a message with the notification text in `msg.payload`:
+
+```javascript
+msg.payload = "Battery voltage is low";
+return msg;
+```
+
+### Overriding Configuration
+
+You can override the configured type and title by setting properties on the incoming message. The type can be specified as either a number (0, 1, 2) or a string ("warning", "alarm", "info"/"information"):
```
Chris927, line 108: Why allow both numbers and strings for the type? It would result in easier-to-read code, and simplify logic, if only the string types are supported.

src/nodes/victron-inject.js
```diff
@@ -0,0 +1,97 @@
...
+                fill: statusColor,
+                shape: 'dot',
+                text: statusText
+              })
+
+              msg.notification = {
+                type,
+                title,
+                message,
+                timestamp: new Date().toISOString()
```
Chris927: The timestamp is possibly a bit "late": In a case where `this.client.publish()` takes 1 second to call the callback (possible if system under load), then the timestamp does not represent the moment the notification was requested.

Maybe take the timestamp at the beginning of the `on('input')` callback, like `const now = new Date()`, then use `now` here.

I actually lack context on why there is a timestamp, so maybe I'm wrong.

test/victron-inject-functions.test.js
```diff
@@ -0,0 +1,303 @@
+const { formatNotification, validateNotificationInput } = require('../src/nodes/victron-inject-functions')
+
+describe('victron-inject-functions', () => {
```
Chris927, line 3: It smells like AI helped creating those tests, that's awesome ;)

Chris927 (changes requested):

examples/Inject Notifications.json
dirkjanfaber: Yes, the example may not be completely up to date with my latest code changes. I'll check that.

src/nodes/victron-inject-functions.js
dirkjanfaber: Good suggestion. Somehow had not thought of that.

src/nodes/victron-inject.html
dirkjanfaber: Hm... not sure. I personally like the fact that you can do both. And that you can increase/decrease the level mathematically.

src/nodes/victron-inject.js
dirkjanfaber: I agree that it isn't that useful. I added it so a downstream flow can do something with it, but that feels a bit far fetched. If needed, the user can add it easily enough. I'll remove that line.

test/victron-inject-functions.test.js
dirkjanfaber: Can't disagree on that. :)

src/nodes/victron-inject-functions.js
dirkjanfaber: Fair point. Adjusted that.

examples/Inject Notifications.json
dirkjanfaber: Clearing the notification is done by acknowledging it via the gui of the GX device (currently). But there are some ways to do it via the dbus too (see https://github.com/victronenergy/venus/wiki/dbus#platform). I am tempted to do that in a subsequent issue.

dirkjanfaber: https://github.com/victronenergy/node-red-contrib-victron/issues/347

Chris927 (approved):

Issue #224, opened 2025-03-21
[BUG] Relay control not working on GX


dirkjanfaber:
**Describe the bug**

See https://community.victronenergy.com/t/venus-os-v3-60-46-node-red-relay-control/25822

dirkjanfaber:
In latest release

PR #359, opened 2026-01-12
Fix/issue 224 relay control

dirkjanfaber:
Fix getRelayServices() to return communityTag structure (issue #224)


Chris927 (approved):
Tested and confirmed working.

PR #361, opened 2026-01-13
switch name and group, sync from dbus cache

Chris927:
name and group of a switch might have been changed outside of Node-RED. In this case, we have the old values in the Node-RED flow, so we retrieve the updated values from the cache we maintain of dbus values.

### DIFF
--- a/src/nodes/victron-virtual-functions.js
+++ b/src/nodes/victron-virtual-functions.js
@@ -453,6 +453,26 @@ function renderDropdownLabels (context) {
   }
 }
 
+export function fetchSwitchNodeNameAndGroupFromCache (id) {
+  if (!id) {
+    return Promise.reject(new Error('id is required'))
+  }
+
+  return fetch(`/victron/cache?filter_by_serial=${id}`)
+    .then(response => response.json())
+    .then(data => {
+      for (const key in data) {
+        if (key.match(/^com.victronenergy\./) && data[key]['/Serial'] === id) {
+          const name = data[key]['/SwitchableOutput/output_1/Settings/CustomName']
+          const group = data[key]['/SwitchableOutput/output_1/Settings/Group']
+          return { name, group }
+        }
+      }
+      // No matching entry found
+      return {}
+    })
+}
+
 export function updateSwitchConfig (context) {
   const container = $('#switch-config-container')
   container.empty()
--- a/src/services/utils.js
+++ b/src/services/utils.js
@@ -284,6 +284,25 @@ function mapCacheToJsonResponse (cache) {
 }
 
 /**
+ * Filter cache, so that the result can be limited for efficiency.
+ * For now, the only filter supported is by device serial.
+ */
+function filterCache (cache, options = {}) {
+  const { filterBySerial } = options
+  if (!filterBySerial) {
+    return cache
+  }
+
+  const filteredCache = {}
+  for (const [device, paths] of Object.entries(cache)) {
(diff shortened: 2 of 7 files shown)


src/services/utils.js
```diff
@@ -283,6 +283,26 @@ function mapCacheToJsonResponse (cache) {
...
+ */
+function filterCache (cache, options = {}) {
+  const { filterBySerial } = options
+  if (!filterBySerial) {
+    return cache
+  }
+
+  const filteredCache = {}
+  for (const [device, paths] of Object.entries(cache)) {
+    console.log('Filtering device:', device, 'by serial:', filterBySerial, 'paths:', paths)
```
dirkjanfaber: Don't think that `console.log` needs to be in production.

src/nodes/victron-virtual-functions.js
```diff
@@ -453,6 +453,29 @@ function renderDropdownLabels (context) {
   }
 }
 
+export function fetchSwitchNodeNameAndGroupFromCache (id) {
+  if (!id) {
+    throw new Error('id is required')
```
dirkjanfaber: Shouldn't this be return Promise.reject(new Error('id is required')) instead?

src/nodes/victron-virtual-functions.js
```diff
@@ -453,6 +453,29 @@ function renderDropdownLabels (context) {
...
+  }
+
+  return fetch(`/victron/cache?filter_by_serial=${id}`)
+    .then(response => response.json())
+    .then(data => {
+      for (const key in data) {
+        if (key.match(/^com.victronenergy\./) && data[key]['/Serial'] === id) {
```
dirkjanfaber, line 465: The /Serial check is a bit double as the cache is already filtered on that. So it should always pass true. Otoh it isn't a big thing to check twice. So fine with leaving this in.

src/services/virtual-switch.js
```diff
@@ -67,6 +67,11 @@ function createSwitchProperties (config, ifaceDesc, iface) {
     persist: true
   }
 
+  ifaceDesc.properties.DeviceInstance = {
+    type: 'i'
+  }
+  ifaceDesc.properties.Serial = { type: 's', persist: true }
```
dirkjanfaber: If needed I am fine with having it in. but it seems a bit unrelated to the other changes.

src/nodes/victron-virtual-functions.js
Chris927: Practically either should be fine, but your version ensures that there is no synchronous exception.

I'll use your version.

src/services/utils.js
Chris927: Will remove.

src/nodes/victron-virtual-functions.js
Chris927: Alright, I'll leave it in.

src/services/virtual-switch.js
Chris927: My plan was to put this into a different commit and PR, but then it slipped through. 🙈

I'll remove it from this PR, and create another (simple one) for this change.

src/nodes/victron-virtual-functions.js
Chris927: FYI, I'm also going to remove the `.catch()` from this function: All it does is log the error, and we log the error anyway (again) when we call the function from the browser code, so it is unnecessary.

PR #362, opened 2026-01-14
Add path documentation for virtual device types

dirkjanfaber:
Show available dbus paths in edit dialog for all virtual device types.

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -957,8 +957,8 @@
       $('#switch-docs-container').empty();
 
       // Show device-specific documentation if available
-      if (DEVICE_TYPE_DOCS[selected]) {
-        const doc = DEVICE_TYPE_DOCS[selected];
+      const doc = DEVICE_TYPE_DOCS[selected];
+      if (doc) {
         const docRow = $(`
         <div class="form-row">
           <div id="device-doc-row" class="victron-doc-box">
--- a/src/nodes/victron-virtual-functions.js
+++ b/src/nodes/victron-virtual-functions.js
@@ -829,8 +829,8 @@ export function checkSelectedVirtualDevice (context) {
     $('#switch-docs-container').empty()
 
     // Show device-specific documentation if available
-    if (DEVICE_TYPE_DOCS[selected]) {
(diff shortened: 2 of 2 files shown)


resources/victron-virtual-functions.js
```diff
@@ -320,6 +320,299 @@
...
+  // Device type documentation
+  const DEVICE_TYPE_DOCS = {
+    acload: {
+      label: 'AC Load',
+      text: `
+    ${INPUT_DOCS}
+    <div>
+      <div><strong>Most relevant paths:</strong>
+        <ul>
+          <li><code>/Ac/{line}/Power</code> &mdash; Power per phase in watts, where <code>{line}</code> is <code>L1</code>, <code>L2</code>, or <code>L3</code>.</li>
```
Chris927, line 332: Having `&mdash;` in the hand-written code is pretty ugly. But, as discussed, we want to find a way to generate this html code from a more concise representation eventually. I suggest to leave it like this until then.

resources/victron-virtual-functions.js
```diff
@@ -662,6 +955,21 @@
     } else {
       $('#default-values-container').show();
       $('#switch-docs-container').empty();
+
+      // Show device-specific documentation if available
+      if (DEVICE_TYPE_DOCS[selected]) {
+        const doc = DEVICE_TYPE_DOCS[selected];
```
Chris927: Have `const doc = ...` before the `if`, then say `if (doc) { ... }`?

src/nodes/victron-virtual-functions.js
```diff
@@ -534,6 +827,21 @@ export function checkSelectedVirtualDevice (context) {
   } else {
     $('#default-values-container').show()
     $('#switch-docs-container').empty()
+
+    // Show device-specific documentation if available
+    if (DEVICE_TYPE_DOCS[selected]) {
+      const doc = DEVICE_TYPE_DOCS[selected]
```
Chris927: same as above

resources/victron-virtual-functions.js
dirkjanfaber: Fully agree, but marking this as resolved for now.

PR #363, opened 2026-01-19
Add linting on git commit

Chris927:
Add [husky](https://typicode.github.io/husky/) to utilize git hooks to run the linter before committing.

Only works if/when developer does an `npm install` after this change.

@dirkjanfaber and I observed various times that, even though we *think* we used the linter, we ended up with diffs that had whitespace changes in unintended locations. Adding linting (and also building derived code via rollup etc, and unit-testing) automatically with each commit should help.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+npm run build
+npm run build:test
+npm run lint
+npm test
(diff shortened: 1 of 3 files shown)


dirkjanfaber (approved):
Looks good to me.

PR #364, opened 2026-01-19
cleanup how we determine name/group properties for switches

Chris927:
This should not change any logic, but determines properties of switches more cleanly.


dirkjanfaber (approved):
Looks good to me.

PR #370, opened 2026-01-22
feat(input-nodes): Add conditional mode for input nodes

dirkjanfaber:
This is ready for review. I have some minor doubts on the UX side still. So curious about what you think of that too. 

Once this looks good, I guess it still needs some extra documentation and/or examples to be updated.

### DIFF
--- a/resources/victron-common.js
+++ b/resources/victron-common.js
@@ -35,11 +35,172 @@ var victronCommon = (function (exports) {
     });
   }
 
+  /**
+   * Filter services based on blacklist patterns
+   * @param {Array} services - Array of service objects
+   * @returns {Array} Filtered services
+   */
+  function filterBlacklistedServices (services) {
+    const blacklist = [
+      /^com\.victronenergy\.adc/,
+      /^com\.victronenergy\.ble/,
+      /^com\.victronenergy\.fronius/,
+      /^com\.victronenergy\.modbusclient\.tcp/,
+      /^com\.victronenergy\.shelly/,
+      /^com\.victronenergy\.vecan\./
+    ];
+
+    return services.filter(function (service) {
+      return !blacklist.some(function (pattern) {
+        return pattern.test(service.service)
+      })
+    })
+  }
+
+  /**
+   * Filter paths to only numeric types (for conditional comparisons)
+   * @param {Array} paths - Array of path objects
+   * @returns {Array} Filtered paths
+   */
+  function filterNumericPaths (paths) {
+    return paths.filter(function (path) {
+      return path.type === 'float' ||
+             path.type === 'integer' ||
+             path.type === 'boolean' ||
+             path.type === 'number'
+    })
+  }
+
+  /**
+   * Count services by name to determine if device instance should be shown
+   * @param {Array} services - Array of service objects
+   * @returns {Object} Map of service name to count
+   */
+  function countServicesByName (services) {
+    const counts = {};
+    services.forEach(function (service) {
+      counts[service.name] = (counts[service.name] || 0) + 1;
+    });
+    return counts
+  }
+
+  /**
+   * Get friendly service type name from D-Bus service path
+   * @param {string} servicePath - D-Bus service path (e.g., "com.victronenergy.tank.socketcan_can0_vi0_uc6536")
+   * @returns {string} Friendly service type name (e.g., "Tank Sensor")
+   */
+  function getServiceTypeName (servicePath) {
+    // Service type mapping (matches registerInputNode calls)
+    const serviceTypeMap = {
+      accharger: 'AC Charger',
+      acload: 'AC Load',
+      acsystem: 'AC System',
+      alternator: 'Alternator',
+      battery: 'Battery Monitor',
+      dcdc: 'DC-DC',
+      dcgenset: 'DC Generator',
+      dcload: 'DC Load',
+      dcsource: 'DC Source',
+      dcsystem: 'DC System',
+      dess: 'Dynamic ESS',
+      digitalinput: 'Digital Input',
+      ess: 'ESS',
+      evcharger: 'EV Charger',
+      fuelcell: 'Fuel Cell',
+      generator: 'Generator',
+      gps: 'GPS',
+      grid: 'Grid Meter',
+      gridmeter: 'Grid Meter',
+      inverter: 'Inverter',
+      meteo: 'Meteo',
+      motordrive: 'Motor Drive',
+      multi: 'Multi RS',
+      pulsemeter: 'Pulse Meter',
+      pump: 'Pump',
+      pvinverter: 'PV Inverter',
+      relay: 'Relay',
+      settings: 'Settings',
+      solarcharger: 'Solar Charger',
+      switch: 'Switch',
+      system: 'System',
+      tank: 'Tank Sensor',
+      temperature: 'Temperature Sensor',
+      vebus: 'VE.Bus System'
+    };
+
+    // Extract service type from path: com.victronenergy.{type}/{instance}
+    const match = servicePath.match(/^com\.victronenergy\.([^./]+)/);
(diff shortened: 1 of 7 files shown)


resources/victron-common.js
```diff
@@ -35,11 +35,172 @@ var victronCommon = (function (exports) {
...
+  }
+
+  /**
+   * Get friendly service type name from D-Bus service path
+   * @param {string} servicePath - D-Bus service path (e.g., "com.victronenergy.tank.socketcan_can0_vi0_uc6536")
+   * @returns {string} Friendly service type name (e.g., "Tank Sensor")
+   */
+  function getServiceTypeName (servicePath) {
+    // Service type mapping (matches registerInputNode calls)
+    const serviceTypeMap = {
```
Chris927, line 94: We have data about services per type already in `src/services/services.json`, we also have `src/nodes/victron-virtual-functions.js` which has a key `label` for each service type. This looks like we yet another place where we have data that describes services.

It would be easier to maintain, if we kept this "meta data" in one place, rather. Is it an option to maintain a "friendly service name" in `src/services/services.json`?

Further, I noticed that `e-drive` is missing, is this on purpose?

src/nodes/victron-nodes.html
```diff
@@ -72,6 +72,160 @@
...
+                <i class="fa fa-filter"></i> Conditional Mode
+                <i class="fa fa-info-circle tooltip-icon" data-tooltip="Enables a second output for conditional evaluation. Output 1 continues to send the raw measurement value (subject to rounding/rate limit settings above). Output 2 sends TRUE/FALSE based on conditions below."></i>
+            </label>
+            <input type="checkbox" id="node-input-conditionalMode">
+        </div>
+
+        <!-- Live Preview Panel -->
+        <div id="conditional-preview-panel" class="victron-preview-panel hidden">
+            <div class="victron-preview-header">
+                <i class="fa fa-eye"></i> Live Preview
```
Chris927, line 89: It's called "Live Preview", but it's not really live:

When the input value changes in dbus (e.g. the state of a switch), the live preview doesn't reflect that.

I suggest to rename it to "Preview" instead of "Live Preview", to not create incorrect expectations.

src/nodes/victron-common.js
```diff
@@ -32,7 +32,163 @@ export function initializeTooltips () {
...
+
+/**
+ * Get friendly service type name from D-Bus service path
+ * @param {string} servicePath - D-Bus service path (e.g., "com.victronenergy.tank.socketcan_can0_vi0_uc6536")
+ * @returns {string} Friendly service type name (e.g., "Tank Sensor")
+ */
+export function getServiceTypeName (servicePath) {
+  // Service type mapping (matches registerInputNode calls)
+  const serviceTypeMap = {
+    accharger: 'AC Charger',
```
Chris927, line 92: ... I should have commented here, not on `resources/victron-common.js` 🙈

src/nodes/victron-nodes.html
```diff
@@ -127,6 +281,422 @@
...
+                                serviceMap[dedupeKey] = service;
+                            }
+                        });
+                    }
+                });
+
+                var services = Object.values(serviceMap);
+
+                // Apply blacklist filter
+                if (window.__victronCommon && window.__victronCommon.filterBlacklistedServices) {
```
Chris927, line 319: It is repetitive to check for the existence of `window.__victronCommon` over and over.

On investigating this further: `victronCommon` is defined globally, because `./resources/victron-common.js` defines it (and because victron-common.js is loaded as a script here).

Thus, we should simplify this to just calling `victronCommon.filterBlacklistedServices()`, without checking if `window.__victronCommon` exists etc.

src/nodes/victron-nodes.html
```diff
@@ -127,6 +281,422 @@
...
+                    services = window.__victronCommon.filterBlacklistedServices(services);
+                }
+
+                if (window.__victronCommon && window.__victronCommon.buildGroupedServiceDropdown) {
+                    try {
+                        window.__victronCommon.buildGroupedServiceDropdown(services, $condition2Service);
+                    } catch (error) {
+                        console.error('Error building grouped dropdown:', error);
+                        // Fallback to old method
```
Chris927, line 328: Why is there a fallback? If there are failure scenarios, they could be handled inside of `buildGroupedServiceDropdown()`.

src/nodes/victron-nodes.html
```diff
@@ -127,6 +281,422 @@
...
+                            }
+                            var $option = $('<option/>')
+                                .val(service.service)
+                                .text(optionText)
+                                .data(service);
+                            $condition2Service.append($option);
+                        });
+                    }
+                } else {
+                    // Fallback if function not available
```
Chris927, line 345: This should be totally unnecessary: If the function isn't available, then we have packaged the files incorrectly and it should blow up with an exception.

src/nodes/victron-nodes.html
```diff
@@ -127,6 +281,422 @@
...
+            });
+        }
+
+        // Update condition 2 paths based on selected service
+        function updateCondition2Paths() {
+            var serviceData = $condition2Service.find(':selected').data();
+            if (serviceData && serviceData.paths) {
+                var paths = serviceData.paths;
+
+                // Filter to numeric types only
```
Chris927, line 377: Why numeric only? Why not allow a string comparison ` === "enabled"` or ` <= "sortKey-123"`?

src/nodes/victron-nodes.html
```diff
@@ -127,6 +281,422 @@
 
     var serviceRoot = (RED.settings.httpNodeRoot || RED.settings.httpAdminRoot || "").replace(/\/$/, "") + '/victron/services/';
 
+    function initializeConditionalMode(node) {
+        var $conditionalMode = $('#node-input-conditionalMode');
```
Chris927, line 285: Why use `var`? We should almost never use `var`, see [this SO answer](https://stackoverflow.com/a/11444416). It's probably safe to replace with `const`.

src/nodes/victron-nodes.html
dirkjanfaber: image

src/nodes/victron-nodes.html
dirkjanfaber: First we do only numeric. If the need arrises we will add other value types.

resources/victron-common.js
dirkjanfaber: Unfortunately there is no 1-1 connection between the dbus services and the `services.json` file, which makes it non trivial to add the labels to that file. But as it is definitely easier to maintain in a `.json` file, moved this to its separate `service-meta.json` file, which deliberately got this name to be ready for future additions.
(see commit b32c95a)

src/nodes/victron-nodes.html
dirkjanfaber: Resolved in commit b32c95a

src/nodes/victron-common.js
dirkjanfaber: This one, and the other issues mentioned below, have been taken care of in commit b32c95a. So marking all of the things as resolved for now and open for a new review.

src/nodes/victron-nodes.html
```diff
@@ -475,25 +513,28 @@
...
+                validateThreshold(); // Will hide warning when conditional mode is disabled
+                stopLivePreview();
             }
         });
 
         // Fetch new data when a service or path is changed
         $('#node-input-service, #node-input-path, #node-input-condition2-service, #node-input-condition2-path').on('change', function() {
-            fetchPreviewCache();
+            if ($conditionalMode.is(':checked')) {
+                fetchPreviewCache();
```
Chris927, line 524: This works for "Custom Input" nodes, but not "Custom Control" nodes. What is the intention?

PR #371, opened 2026-01-28
Add lint-staged to build ./resources via pre-commit hook (via husky)

Chris927:
(below is from the commit message)

Noteworthy about the configuration:

- lint-staged does not handle the case easily where a command not only changes the git-staged file but also other files. Our 'npm run build' does that, though. Therefore we need to configure lint-staged via anonymous functions in lint-staged.config.js. See
https://github.com/lint-staged/lint-staged/issues/1149 for details.

- To avoid accidental, manual changes in ./resources, we check for such changes in .husky/pre-commit. Then, to circumvent lint-staged's limitations, we call 'git add' ourselves in .husky/pre-commit.


package-lock.json
```diff
@@ -64,6 +65,7 @@
       "integrity": "sha512-H3mcG6ZDLTlYfaSNi0iOKkigqMFvkTKlGUYlD8GW7nNOYRrevuA46iTypPyv+06V3fEmvvazfntkBU34L0azAw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
```
dirkjanfaber: Noticing this line more often. And recall that it has been on and off a couple of times in the file. According to https://github.com/npm/cli/issues/8690 that seems deprecated in newer versions of npm.

PR #379, opened 2026-02-09
fix(notification): Accept string numbers for notification type

dirkjanfaber:
Refactor validateNotificationInput to correctly parse string representations of 0, 1, and 2 as valid notification types. Updated unit tests to use a parameterized test for these cases.


Chris927 (approved):
Tested and reviewed, I approve.

PR #386, opened 2026-02-23
Refactor/virtual device types

dirkjanfaber:
A fresh attempt for the refactoring. Which turned out to be more easy than the previous attempts. 
So pulled the individual virtual device types in separate files for more easy maintainability.

### DIFF
--- a/src/nodes/victron-virtual/device-type/battery.js
+++ b/src/nodes/victron-virtual/device-type/battery.js
@@ -1,5 +1,5 @@
 const properties = {
-  Capacity: { type: 'd', format: (v) => v != null ? v.toFixed(0) + 'Ah' : '', persist: true },
+  Capacity: { type: 'd', format: (v) => v != null ? v.toFixed(0) + 'Ah' : '', persist: false },
   'Dc/0/Current': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'A' : '', immediate: true },
   'Dc/0/Power': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'W' : '', immediate: true },
   'Dc/0/Voltage': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'V' : '', immediate: true },
--- a/src/nodes/victron-virtual/device-type/tank.js
+++ b/src/nodes/victron-virtual/device-type/tank.js
@@ -27,7 +27,7 @@ const properties = {
       11: 'Raw water'
     }[v] || 'unknown'),
     value: 0,
-    persist: true
+    persist: false
   },
   Level: { type: 'd', format: (v) => v != null ? v.toFixed(0) + '%' : '', persist: 60 },
   RawUnit: { type: 's', persist: true },


test/victron-virtual-device-types.test.js
```diff
@@ -0,0 +1,33 @@
...
+const deviceTypeDir = path.join(__dirname, '../src/nodes/victron-virtual/device-type')
+
+const moduleFiles = fs.readdirSync(deviceTypeDir)
+  .filter(f => f.endsWith('.js'))
+  .map(f => ({ name: path.basename(f, '.js'), mod: require(path.join(deviceTypeDir, f)) }))
+
+describe('virtual device type module contracts', () => {
+  test.each(moduleFiles)('$name exports properties as an object', ({ mod }) => {
+    expect(mod.properties).toBeDefined()
+    expect(typeof mod.properties).toBe('object')
```
Chris927, line 15: When inspecting test coverage, I found we can increase coverage by running (most) of the `format()` functions here, to ensure they produce a string, see diff below. Maybe either add this to this PR, or we can do a follow-up PR with this:

```patch
diff --git a/test/victron-virtual-device-types.test.js b/test/victron-virtual-device-types.test.js
index 6f414f5..57688f0 100644
--- a/test/victron-virtual-device-types.test.js
+++ b/test/victron-virtual-device-types.test.js
@@ -9,11 +9,37 @@ const moduleFiles = fs.readdirSync(deviceTypeDir)
   .filter(f => f.endsWith('.js'))
   .map(f => ({ name: path.basename(f, '.js'), mod: require(path.join(deviceTypeDir, f)) }))

+function buildRandomPropertyValue(type) {
+  switch (type) {
+    case 'd': return Math.random() * 100 // random number for 'd' type
+    case 's': return 'test string' // fixed string for 's' type
+    case 'b': return Math.random() < 0.5 // random boolean for 'b' type
+    default: return null // default to null for unknown types
+  }
+}
+
+function assertFormatFunctions(properties) {
+  for (const key in properties) {
+    const prop = properties[key]
+    if (prop.format) {
+
+      // Test that the format function returns a string for valid input
+      const testValue = buildRandomPropertyValue(prop.type)
+      const formatted = prop.format(testValue)
+      expect(typeof formatted).toBe('string')
+    }
+  }
+}
+
 describe('virtual device type module contracts', () => {
   test.each(moduleFiles)('$name exports properties as an object', ({ mod }) => {
     expect(mod.properties).toBeDefined()
     expect(typeof mod.properties).toBe('object')
     expect(Array.isArray(mod.properties)).toBe(false)
+
+    // assert that format functions, if present, produce a string output, given valid input
+    assertFormatFunctions(mod.properties)
+
   })

   test.each(moduleFiles)('$name exports initialize as a function', ({ mod }) => {
```

src/nodes/victron-virtual/index.js
```diff
@@ -0,0 +1,790 @@
...
+    }
+  }
+
+  const result = {
+    emit: function () {
+    }
+  }
+
+  for (const key in properties[dev]) {
+    const propertyValue = JSON.parse(JSON.stringify(properties[dev][key]))
```
Chris927, line 145: Why `stringify()` and then `parse()`? If it is necessary to do this, please add a comment.

src/nodes/victron-virtual/index.js
```diff
@@ -0,0 +1,790 @@
...
+
+  result.DeviceInstance = { type: 'i' }
+  result.CustomName = { type: 's', persist: true }
+  result.Serial = { type: 's', persist: true }
+
+  return result
+}
+
+function getIface (dev) {
+  const actualDev = dev === 'generator' ? 'genset' : dev === 'e-drive' ? 'motordrive' : dev
```
Chris927, line 131: Dubious usage of `actualDev` vs `dev`: We use `actualDev` in the next line, but further down we use `dev`. Why?

This may be a bug, or if not, there should be a comment explaining why we distinguish `dev` vs `actualDev`.

src/nodes/victron-virtual/index.js
```diff
@@ -0,0 +1,790 @@
...
+  if (!properties[actualDev]) {
+    return {
+      emit: function () {
+      }
+    }
+  }
+
+  const result = {
```
Chris927, line 139: This object definition duplicates the object we return within the previous `if()`. I suggest to refactor this, so that the `for (const key in ...)` below iterates zero times if there are no properties. The return value at the end of the function will be the same (as the object returned in the previous `if()`).

src/nodes/victron-virtual/index.js
```diff
@@ -0,0 +1,790 @@
...
+      } catch (err) {
+        console.error(`Failed to load virtual device type "${name}":`, err)
+      }
+    })
+} catch (err) {
+  console.error('Failed to load virtual device types:', err)
+}
+
+function getIfaceDesc (dev) {
+  const actualDev = dev === 'generator' ? 'genset' : dev === 'e-drive' ? 'motordrive' : dev
```
Chris927, line 108: See comment below concerning `actualDev` vs `dev`.

test/victron-virtual-device-types.test.js
dirkjanfaber: Good idea, but I propose doing this in a follow-up PR.

src/nodes/victron-virtual/index.js
dirkjanfaber: This is also a migration from the [orginal](https://github.com/victronenergy/node-red-contrib-victron/blob/a16350faa76e7872fb5e4320cd5a0a8852bd40d1/src/nodes/victron-virtual.js#L323). 

And this should get reviewed and changed via PR #398 (which it doesn't right now).

src/nodes/victron-virtual/index.js
dirkjanfaber: These are one-on-one migrations of the [original code](https://github.com/victronenergy/node-red-contrib-victron/blob/a16350faa76e7872fb5e4320cd5a0a8852bd40d1/src/nodes/victron-virtual.js#L337).
So I agree that this is dubious and needs a closer look. But for this PR I consider it not that relevant.

PR #389, opened 2026-02-25
Feature/virtual device presence

dirkjanfaber:


test/victron-virtual-presence.test.js
```diff
@@ -0,0 +1,323 @@
...
+      }
+    }
+    return node
+  }
+
+  const text = 'Virtual battery'
+  const iface = { DeviceInstance: 100 }
+
+  function makeSetPresence (node) {
+    return function setPresence (connected, onDone) {
```
Chris927: This whole test is pointless, as it doesn't require/import the unit under test.

Please revise, and ensure the actual unit under test is tested, not a function that repeats what the actual unit under test is doing.

src/nodes/victron-virtual/index.js
```diff
@@ -212,13 +213,23 @@ module.exports = function (RED) {
...
+      }
       const outputs = [msg]
       // Fill remaining outputs with null
       for (let i = 1; i < config.outputs; i++) {
         outputs.push(null)
       }
       node.send(outputs)
 
+      if (userSetConnected) {
+        const targetConnected = msg.connected === 'toggle' ? !node.presenceConnected : !!msg.connected
```
Chris927: I suggest to not support different ways to set the virtual device as connected: Rather provide `connected = true` and `connected = false`, but not `'toggle'`: With true/false only, the solution is simpler, and should still allow users to code whatever they need in terms of presence. In addition, the "toggle" implementation assumes "exactly once" delivery of messages, which will not always be the case.

test/victron-virtual-presence.test.js
dirkjanfaber: Resolved by pulling out the function into a `helper.js` and using that function in the test.

src/nodes/victron-virtual/index.js
dirkjanfaber: Removed this functionality + the documentation of the toggling.

PR #390, opened 2026-02-25
fix(conditional): merge services when determining dropdowns

Chris927:
fix(conditional): merge services when determining dropdowns for second condition

This aims at achieving what [PR#383](https://github.com/victronenergy/node-red-contrib-victron/pull/383) did, but less hacky, and with a lengthy comment explaining what we are doing. This PR is the result of yesterday's discussion between @dirkjanfaber and me.

This could possibly be simplified further, but we should simplify at a later stage (when we have (better) test coverage of the relevant logic).

### DIFF
--- a/src/nodes/victron-nodes.html
+++ b/src/nodes/victron-nodes.html
@@ -304,13 +304,39 @@
                 victronCommon.setServiceLabels(allServices.serviceLabels);
 
                 // Combine all input services into a single array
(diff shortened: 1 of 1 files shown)


dirkjanfaber (approved):
Works perfect. Marked this as 3.70 because it is a possible backport candidate.

PR #394, opened 2026-02-26
fix(virtual): device class for e-drive and generator

citolen:
Closes https://github.com/victronenergy/node-red-contrib-victron/issues/393

PR #398, opened 2026-03-04
fix(virtual): correct device class for e-drive and generator

### LINKED ISSUE #393: Bug: wrong device class for E-drives and generators
**Describe the bug**
Virtual E-drives and generators are using the values `e-drive` and `generator` as the class in `ClassAndVrmInstance`.

**To Reproduce**
Create a virtual E-drive or generator and check the `ClassAndVrmInstance` value in the localsettings.

**Expected behavior**
E-drive should use the device class `motordrive`.
Generator should use the device class `dcgenset` or `genset` depending on its type.

dirkjanfaber:
Closes #393
Supersedes #394 (based on @citolen's work — thank you Nils!)

## What changed

Virtual e-drive and generator nodes were storing the wrong class in `ClassAndVrmInstance` in localsettings (using the UI device type `e-drive`/`generator` instead of the correct D-Bus types
`motordrive`/`genset`/`dcgenset`).

## Commits

1. **fix(virtual): device class for e-drive and generator** *(Nils Citoleux)*
   Extract `getDeviceTypeForConfig()` helper and use it for the `ClassAndVrmInstance` default so newly created devices get the correct class from the start.

2. **fix(virtual): lock generator AC/DC type after deployment**
   Disable the generator type (AC/DC) selector in the edit dialog once  the node has been deployed, preventing a class mismatch between localsettings and the running D-Bus service.

3. **fix(virtual): migrate legacy ClassAndVrmInstance on redeploy**
   On startup, detect legacy `generator:N` / `e-drive:N` values returned by `AddSettings` and call `SetValue` to overwrite them with the correct D-Bus type while preserving the VRM instance number.
   Also updates the now-stale comment in `virtual-device-cleanup.js`.

### DIFF
--- a/src/nodes/victron-virtual/index.js
+++ b/src/nodes/victron-virtual/index.js
@@ -104,8 +104,13 @@ try {
   console.error('Failed to load virtual device types:', err)
 }
 
-function getIfaceDesc (dev) {
-  const actualDev = dev === 'generator' ? 'genset' : dev === 'e-drive' ? 'motordrive' : dev
+function getActualDeviceType (type, subtype) {
+  if (type === 'generator') return subtype === 'dc' ? 'dcgenset' : 'genset'
+  if (type === 'e-drive') return 'motordrive'
+  return type
+}
+
+function getIfaceDesc (actualDev) {
   if (!properties[actualDev]) {
     return {}
   }
@@ -113,7 +118,7 @@ function getIfaceDesc (dev) {
   const result = {}
 
   // Deep copy the properties, including format functions
-  for (const [key, value] of Object.entries(properties[dev])) {
+  for (const [key, value] of Object.entries(properties[actualDev])) {
     result[key] = { ...value }
     if (typeof value.format === 'function') {
       result[key].format = value.format
@@ -127,8 +132,7 @@ function getIfaceDesc (dev) {
   return result
 }
 
-function getIface (dev) {
-  const actualDev = dev === 'generator' ? 'genset' : dev === 'e-drive' ? 'motordrive' : dev
+function getIface (actualDev) {
   if (!properties[actualDev]) {
     return {
       emit: function () {
@@ -141,8 +145,8 @@ function getIface (dev) {
     }
   }
 
-  for (const key in properties[dev]) {
-    const propertyValue = JSON.parse(JSON.stringify(properties[dev][key]))
+  for (const key in properties[actualDev]) {
+    const propertyValue = JSON.parse(JSON.stringify(properties[actualDev][key]))
 
     if (propertyValue.value !== undefined) {
       result[key] = propertyValue.value
@@ -381,11 +385,7 @@ module.exports = function (RED) {
         return
       }
 
-      const actualDeviceType = config.device === 'generator'
-        ? (config.generator_type === 'dc' ? 'dcgenset' : 'genset')
-        : config.device === 'e-drive'
-          ? 'motordrive'
-          : config.device
+      const actualDeviceType = getActualDeviceType(config.device, config.generator_type)
 
       const serviceName = `com.victronenergy.${actualDeviceType}.virtual_${self.id}`
       const interfaceName = serviceName
@@ -581,10 +581,12 @@ module.exports = function (RED) {
           const parts = currentClassAndVrmInstance.split(':')
           const currentClass = parts[0]
           const vrmInstance = parts[1]
-          if (currentClass !== actualDeviceType && vrmInstance != null) {
+          if (currentClass !== actualDeviceType && !vrmInstance) {
+            throw new Error(`Invalid ClassAndVrmInstance value: ${currentClassAndVrmInstance}`)
+          }
+          if (currentClass !== actualDeviceType) {
             const newValue = `${actualDeviceType}:${vrmInstance}`
-            let migrationSucceeded = false
-            await new Promise(resolve => {
+            const migrationSucceeded = await new Promise(resolve => {
               usedBus.invoke({
                 path: `/Settings/Devices/virtual_${node.id}/ClassAndVrmInstance`,
                 destination: 'com.victronenergy.settings',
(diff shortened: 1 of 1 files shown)


src/nodes/victron-virtual.js
```diff
@@ -1159,6 +1166,35 @@ module.exports = function (RED) {
           return null
         }
         iface.DeviceInstance = getDeviceInstance(settingsResult)
+
+        // Migrate legacy ClassAndVrmInstance values ('generator', 'e-drive') to the correct
+        // D-Bus type. AddSettings only sets defaults so existing values are never overwritten
+        // automatically — we need an explicit SetValue to fix them.
+        const currentClassAndVrmInstance = settingsResult?.[0]?.[2]?.[1]?.[1]?.[0] || settingsResult?.[1]?.[0]
+        if (currentClassAndVrmInstance) {
```
citolen: I think there is a possibility for `iface.DeviceInstance` to be out of sync.
When migrating to the correct device class, the original instance id might already be taken and localsettings will pick a free instance id.

The migration logic could be done before the assignment of `iface.DeviceInstance` or it can be assigned a second time using the result from the migrating SetValue call. 

Example:
Let's say you have `e-drive:100` and `motordrive:100` co-existing.
During the migration of `e-drive:100`, localsettings will return `101`.
`iface.DeviceInstance` would still be `100`.

dirkjanfaber: Good catch. The migration now re-reads the actual `ClassAndVrmInstance` after `SetValue` by calling `AddSettings` again. Since AddSettings returns the actually-assigned value (including any conflict-resolved instance), `iface.DeviceInstance` is updated from that result rather than from the pre-migration settingsResult.

The updated flow:
  1. AddSettings('genset:100') -> returns e.g. generator:5 (existing legacy value)
  2. iface.DeviceInstance = 5 (from legacy value)
  3. Migration: SetValue('genset:5')
  4. Re-read: AddSettings('genset:5') -> returns actual assigned value (e.g. genset:6 if 5 was taken)
  5. iface.DeviceInstance updated from re-read result
  
Note that I rebased all of it on an updated master (which has the virtual devices refactored into separate files for ease of maintainability). So you'd want to check the `src/nodes/victron-virtual/index.js` file now.

src/nodes/victron-virtual/index.js
```diff
@@ -572,6 +572,53 @@ module.exports = function (RED) {
...
+        // automatically — we need an explicit SetValue to fix them.
+        const currentClassAndVrmInstance = settingsResult?.[0]?.[2]?.[1]?.[1]?.[0] || settingsResult?.[1]?.[0]
+        if (currentClassAndVrmInstance) {
+          const parts = currentClassAndVrmInstance.split(':')
+          const currentClass = parts[0]
+          const vrmInstance = parts[1]
+          if (currentClass !== actualDeviceType && vrmInstance != null) {
+            const newValue = `${actualDeviceType}:${vrmInstance}`
+            let migrationSucceeded = false
+            await new Promise(resolve => {
```
Chris927: It's cleaner to rewrite this to:

```js
const migrationSucceeded = await new Promise(resolve => {
  ...
  if (err) {
    debug(...);
    resolve(false);
  } else {
    debug(...);
    resolve(true);
  }
});
```

src/nodes/victron-virtual/index.js
Chris927, line 387: According the logic below, `generator` maps to `dcgenset` or `genset`, and `e-drive` maps to `motordrive`. We have almost the same logic in lines 108 and 131, but there we ignore `generator_type`.

This *is* related to this PR, as we use the value of `actualDeviceType` in new logic.

I suggest to clean this up now by introducing a function `getActualDeviceType({ device, generator_type })` which we call in all those places.

src/nodes/victron-virtual/index.js
```diff
@@ -572,6 +572,53 @@ module.exports = function (RED) {
...
+
+        // Migrate legacy ClassAndVrmInstance values ('generator', 'e-drive') to the correct
+        // D-Bus type. AddSettings only sets defaults so existing values are never overwritten
+        // automatically — we need an explicit SetValue to fix them.
+        const currentClassAndVrmInstance = settingsResult?.[0]?.[2]?.[1]?.[1]?.[0] || settingsResult?.[1]?.[0]
+        if (currentClassAndVrmInstance) {
+          const parts = currentClassAndVrmInstance.split(':')
+          const currentClass = parts[0]
+          const vrmInstance = parts[1]
+          if (currentClass !== actualDeviceType && vrmInstance != null) {
```
Chris927: In my understanding `vrmInstance` should never be `null`. Rather say `if (!vrmInstance) { throw ... }`?

src/services/virtual-device-cleanup.js
```diff
@@ -39,8 +39,9 @@ function filterInactiveVirtualDevices (deviceEntries, activeServices) {
       const deviceType = classAndVrmInstance.split(':')[0] // e.g., "switch", "generator", etc.
 
       // Map ClassAndVrmInstance device types to DBus service name prefixes.
-      // ClassAndVrmInstance stores the UI device type (config.device), which doesn't always
-      // match the DBus service name directly.
+      // New devices store the D-Bus type directly (e.g. 'genset', 'dcgenset', 'motordrive').
+      // Legacy entries (created before this fix) stored the UI type ('generator', 'e-drive').
+      // Handle both formats for backward compatibility:
```
Chris927, line 44: To confirm I'm understanding this correctly: Based on the migration we are adding in this PR, `generator` and `e-drive` should not appear as values for `deviceType` any more. *However*, they might still occur in settings, if the corresponding virtual device has been deleted before we ran the migration.

If this is correct, then we do need the `possibleServiceNames` based on the "old" deviceTypes, otherwise we wouldn't.

dirkjanfaber:
Updated the PR, resolving all feedback. Ideally I would have put the new `getActualDeviceType` function in a unit test. PR #389 will make that a bit easier. So either we need that one approved first and then update this PR once more or do that after both are merged.

PR #403, opened 2026-03-04
feat(virtual): add virtual pulse meter device type

### LINKED ISSUE #382: [Feature] Virtual IO - e.g. for Pulse Meter
In Venus OS 3.70 i can use the integrated Digital I/O pins as Pulse Meter. 
So far it seems not to be possible to either "set" the inputs via Node-Red or even create virtual inputs that can be set via node-red. 

The thing is that i want to include I/O to be shown in the Dashboard and VRM but not via real IO, but via mqtt and node-red. In my case specially to add the Diesel/Oil consumption of my heating in liters. I have all the data i need via mqtt, i just don't see a way to inject the pulses into the IO Pulsemeter i configured in Venus.

dirkjanfaber:
## Summary

- New `pulsemeter` virtual device exposing `/Count` and `/Aggregate` on D-Bus
- Optional auto-compute: `Aggregate = Count × multiplier` when Count changes (default: 0.001 m³/pulse)
- 2 output ports: Passthrough and Aggregate (emits on every value change)
- Config dialog, full docs, and 100% test coverage on the new module

Closes #382

## Testing

Open the dialog and click Done to activate the 2nd output port. With auto-compute disabled, inject `{ Count: 1000, Aggregate: 1.0 }` and verify both values appear on D-Bus. With auto-compute enabled (multiplier 0.001), inject `{ Count: 1000 }` and verify `/Aggregate` = 1.0 on D-Bus and on output port 1. Also verify that changing `/Count` directly via D-Bus updates `/Aggregate` and emits on port 1.

### DIFF
--- a/src/nodes/victron-virtual/device-type/pulsemeter.js
+++ b/src/nodes/victron-virtual/device-type/pulsemeter.js
@@ -1,0 +1,31 @@
+const properties = {
+  Count: { type: 'i', format: (v) => v != null ? String(v) : '', persist: true, immediate: true },
+  Aggregate: { type: 'd', format: (v) => v != null ? v.toFixed(3) + 'm³' : '', persist: true, immediate: true }
+}
+
+function onPropertyChanged (propName, propValue, iface, config) {
+  if (config.auto_aggregate && propName === 'Count' && propValue != null) {
+    const multiplier = Number(config.pulsemeter_multiplier)
+    if (!isNaN(multiplier) && multiplier > 0) {
+      const aggregate = propValue * multiplier
+      return {
+        setValues: { Aggregate: aggregate },
+        outputIndex: 1,
+        msg: { payload: aggregate, topic: '/Aggregate', source_path: '/Aggregate' }
+      }
+    }
+  }
+
+  if (propName === 'Aggregate' && propValue != null) {
+    return {
+      outputIndex: 1,
+      msg: { payload: propValue, topic: '/Aggregate', source_path: '/Aggregate' }
+    }
+  }
+}
+
+function initialize (config, ifaceDesc, iface, node) {
+  return 'Virtual pulse meter'
+}
+
+module.exports = { properties, initialize, onPropertyChanged, label: 'Pulse meter' }
--- a/src/nodes/victron-virtual/index.js
+++ b/src/nodes/victron-virtual/index.js
@@ -20,6 +20,7 @@ const gpsModule = require('./device-type/gps')
 const gridModule = require('./device-type/grid')
 const meteoModule = require('./device-type/meteo')
 const motordriveModule = require('./device-type/motordrive')
+const pulsemeterModule = require('./device-type/pulsemeter')
 const pvinverterModule = require('./device-type/pvinverter')
 const switchModule = require('./device-type/switch')
 const tankModule = require('./device-type/tank')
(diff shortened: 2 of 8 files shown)


src/nodes/victron-virtual/index.js
```diff
@@ -626,6 +630,11 @@ module.exports = function (RED) {
         // TODO: S2: Should we rename this?
```
Chris927, line 630: Not related to this PR, but I believe this `TODO` comment can be removed.

src/nodes/victron-virtual/device-type/pulsemeter.js
```diff
@@ -0,0 +1,31 @@
+const properties = {
+  Count: { type: 'i', format: (v) => v != null ? String(v) : '', persist: true, immediate: true },
+  Aggregate: { type: 'd', format: (v) => v != null ? v.toFixed(3) + 'm³' : '', persist: true, immediate: true }
+}
+
+function onPropertyChanged (propName, propValue, iface, config) {
```
Chris927, line 6: When having more than 2 or so function arguments, I prefer a single object and destructuring, as it avoid certain possible bugs and is often more readable:

```js
function onPropertyChange({ name, value, iface, config }) {
  ...
}
```

Then we can call it like this:

```js
deviceModule.onPropertyChanged({ name: propName, value: propValue, iface, config })
```

src/nodes/victron-virtual/index.js
```diff
@@ -647,6 +656,32 @@ module.exports = function (RED) {
...
+            return
+          }
+
+          // Allow device modules to react to property changes and compute derived values
+          const deviceModule = deviceModules[config.device]
+          if (deviceModule && typeof deviceModule.onPropertyChanged === 'function' && node.setValuesLocally) {
+            const result = deviceModule.onPropertyChanged(propName, propValue, iface, config)
+            if (result != null) {
+              // Set derived values (e.g. auto-compute Aggregate from Count).
+              // Mark them first so their emitCallback does not re-trigger onPropertyChanged.
```
Chris927, line 672: I suggest to avoid recursion, it obfuscates the possible call paths. We can rather refactor how we actually set values:

1. Define new function `setOwnDbusProperties({ key1: value1, key2: value2 })`:

- calls `setValuesLocally(values, { skipEmitItemsChanged: true })`. `skipEmitItemsChanged` option does not exist yet, we need to add this.
- iterates over the values, and persists them, if necessary.

2. Pass the new function when calling `onPropertyChanged()`:

```js
... deviceModule.onPropertyChanged({
  name: propName,
  value: propValue,
  iface,
  config,
  setOwnDbusProperties
});
```

3. In the pulsemeter implementation, use it:

```js
  ...
  setOwnDbusProperties({ Aggregate: aggregate })
```

This makes the interface for derived values more explicit, and easier to understand: There is no recursion, and we don't need the set `skipOnPropertyChanged` to track.

And by the way, `skipOnPropertyChanged` sounds like it's a boolean, not a set.

Chris927 (changes requested):

Chris927:
I suggest we update this to use the [new onPropertiesChanged](https://github.com/Chris927/dbus-victron-virtual/pull/20) once available.

dirkjanfaber:
Closing in favour of #476

PR #412, opened 2026-03-09
Switch improvements

dirkjanfaber:
After playing around a bit I found some ways to get make the switch controls a bit more intuitive. Split the changes up in 2 commits for both readability and to make it easier to backport dc3db74a4a8819184ae4fc7875d9d46dcce13edf. 

- dc3db74a4a8819184ae4fc7875d9d46dcce13edf - feat(switch): add Auto mode output for three-state and bilge pump - backportable to 1.6 .x. This adds a third output which emits the status of the Auto mode.
- 72bd31c27a29517ea5a4a5ed83c6a79efede2b98 - feat(switch): improve node status display for virtual switch.

### DIFF
--- a/src/nodes/victron-virtual-switch.js
+++ b/src/nodes/victron-virtual-switch.js
@@ -13,8 +13,7 @@ const dbus = require('dbus-native-victron')
 const debug = require('debug')('victron-virtual-switch')
 const debugInput = require('debug')('victron-virtual-switch:input')
 const debugConnection = require('debug')('victron-virtual-switch:connection')
-const { DEBOUNCE_DELAY_MS } = require('./victron-virtual-constants')
-const { validateVirtualDevicePayload, validateLightControls, debounce } = require('../services/utils')
+const { validateVirtualDevicePayload, validateLightControls } = require('../services/utils')
 const { createSwitchProperties, handleSwitchOutputs, updateSwitchStatus, emitInitialSwitchOutputs } = require('../services/virtual-switch')
 const { filterInactiveVirtualDevices } = require('../services/virtual-device-cleanup')
 
@@ -51,31 +50,6 @@ module.exports = function (RED) {
 
     node.retryOnConnectionEnd = true
     node.pendingCallsToSetValuesLocally = []
-
-    const debouncedSetters = new Map()
-
-    function shouldApplyImmediately (key) {
-      // if (node.ifaceDesc && node.ifaceDesc.properties && node.ifaceDesc.properties[key]) {
-      //   return node.ifaceDesc.properties[key].immediate === true
-      // }
-      // return false
-      return true
-    }
-
-    function getDebouncedSetter (key) {
-      if (!debouncedSetters.has(key)) {
-        const setter = debounce((value) => {
-          debugInput(`Applying debounced value for ${key}: ${value}`)
-          try {
-            node.setValuesLocally({ [key]: value })
-          } catch (err) {
-            node.error(`Failed to apply debounced value for ${key}: ${err.message}`)
-          }
-        }, DEBOUNCE_DELAY_MS)
-        debouncedSetters.set(key, setter)
-      }
-      return debouncedSetters.get(key)
-    }
 
     function handleInput (msg, done) {
       // Send passthrough message FIRST, before any validation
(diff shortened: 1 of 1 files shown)


src/nodes/victron-virtual-functions.js
```diff
@@ -168,13 +168,13 @@ export const SWITCH_TYPE_DOCS = {
...
-    '<div><strong>Outputs:</strong><ol><li><code>Passthrough</code> &mdash; Outputs the original <tt>msg.payload</tt> without modification</li><li><code>State</code> &mdash; <tt>msg.payload</tt> contains a <tt>0</tt> or <tt>1</tt> representing the on/off state of the pump</li></ol></div>',
+    '<div><strong>Most relevant path(s):</strong><ul><li><code>/SwitchableOutput/output_1/State</code> &mdash; Requested on/off state of channel, separate from dimming.</li><li><code>/SwitchableOutput/output_1/Auto</code> &mdash; Auto mode (0=Manual, 1=Auto). When in auto mode, the GX device controls the state based on sensor input.</li><li><code>/SwitchableOutput/output_1/Alarm</code> &mdash; Indicates if an alarm condition is present (e.g., high water level).</li><li><code>/SwitchableOutput/output_1/Measurement</code> &mdash; If supported by the connected device, this path may provide additional measurement data, such as water level percentage.</li></ul></div>',
```
Chris927: "dimming" seems to be the wrong wording for a bilge pump.

src/nodes/victron-virtual-switch.js
Chris927: I'm sure you want to apply this here:

```patch
diff --git a/src/nodes/victron-virtual-switch.js b/src/nodes/victron-virtual-switch.js
index c00355a..f74c8d3 100644
--- a/src/nodes/victron-virtual-switch.js
+++ b/src/nodes/victron-virtual-switch.js
@@ -433,6 +433,7 @@ module.exports = function (RED) {
             })
           }

+          updateSwitchStatus(config, node)
           handleSwitchOutputs(config, node, propName, propValue)
         }
```

While testing, it took me a while to figure out why changes in Gui V2 did not reflect ;)

src/nodes/victron-virtual-switch.html
```diff
@@ -43,13 +43,15 @@
...
+                9: 'Auto mode',    // THREE_STATE
+                10: 'Auto mode',   // BILGE_PUMP
                 11: 'LightControls', // RGB_COLOR_WHEEL
                 12: 'LightControls', // CCT_WHEEL
                 13: 'LightControls'  // RGB_WHITE_DIMMER
             }
 
             const outputCounts = {
-                0: 2, 1: 2, 2: 3, 3: 2, 4: 3, 6: 2, 7: 2, 8: 3, 9: 2, 10: 2, 11: 3, 12: 3, 13: 3
+                0: 2, 1: 2, 2: 3, 3: 2, 4: 3, 6: 2, 7: 2, 8: 3, 9: 3, 10: 3, 11: 3, 12: 3, 13: 3
```
Chris927: Virtual three-state switches and virtual bilge pumps created before this change will have 2 outputs, respectively. That's fine, but may be confusing at times. (It confused me.)

If there is an easy way to migrate existing nodes to have 3 outputs, it might be worthwhile doing it. I am not aware of an easy way, though.

src/services/virtual-switch.js
```diff
@@ -499,8 +511,147 @@ function handleSwitchOutputs (config, node, propName, propValue) {
...
+      switchType === SWITCH_TYPE_MAP.RGB_COLOR_WHEEL ||
+      switchType === SWITCH_TYPE_MAP.CCT_WHEEL ||
+      switchType === SWITCH_TYPE_MAP.RGB_WHITE_DIMMER
+    ) {
+      const lightControls = iface['SwitchableOutput/output_1/LightControls']
+      if (lightControls != null) {
+        const topicLabel = SWITCH_THIRD_OUTPUT_LABEL[switchType] || 'lightcontrols'
+        const [hue, saturation, brightness, white, colorTemp] = lightControls
+        const rgb = hsbToRgb(hue, saturation, brightness)
+        outputMsgs[2] = {
```
Chris927: A lot of duplication with line 492. How about a function `buildLightControlOutputMessage()` or similar, which we use in both places? If anything changes concerning any of these properties, we will likely overlook that both parts of the code need to be changed.

src/nodes/victron-virtual-functions.js
dirkjanfaber: I agree, seems to be in from the beginning and I took the text from https://github.com/victronenergy/venus/wiki/dbus#switch.
Going to leave the ", separate from dimming" out of here.

src/nodes/victron-virtual-switch.html
dirkjanfaber: I can't think of an easy way to do this either. If it is possible at all.

dirkjanfaber:
Took care of all of the remarks of the initial review. Updated the commits (and force pushed) to make the history clean again and still split it up in 2 commits.

PR #417, opened 2026-03-17
feat(virtual-switch): add Show in UI option

dirkjanfaber:
Adds a Show in UI section to the virtual switch node edit dialog, controlled by two checkboxes (Local UI / Remote UI) backed by the `Settings/ShowUIControl` bitmask:
  0bxx1 - show in all UIs (both checked -> value 1)
  0bx10 - local UI only (GX device, MFD, LAN WASM)
  0b1x0 - remote UI only (VRM remote console and switch pane)
  0b000 - hidden

Introduces two reusable helpers in victron-virtual-functions.js:
  renderShowInUICheckboxes(containerId, savedValue, targetInputId)
  getShowUIValue(containerId)

These will also be used by the generic input API node.

image

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -750,6 +750,54 @@
     }
   };
 
+  function renderShowInUICheckboxes (containerId, savedValue, targetInputId) {
+    const localId = `${containerId}-local`;
+    const remoteId = `${containerId}-remote`;
+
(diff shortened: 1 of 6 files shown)


Chris927 (changes requested):
As per comment in the conversation.

Chris927 (approved):

PR #430, opened 2026-03-24
fix(virtual-switch): allow array of strings for dropdown Labels

### LINKED ISSUE #429: Unable to set labels on dropdown virtual switches
Injecting the json as an array into a dropdown virtual switch - as per the docs does not work.
It fails with:
Invalid value types for keys: /SwitchableOutput/output_1/Settings/Labels. Expected: string, number, boolean, null, or array of numbers.

As per docs:
/SwitchableOutput/output_1/Settings/Labels — defines the labels as a string array: ['Label 1', 'Label 2', 'Label 3']. Mandatory for this type.

dirkjanfaber:
The validateVirtualDevicePayload function only accepted arrays of numbers, rejecting the string array required by the dropdown virtual switch Labels field.

Allow homogeneous arrays of strings in addition to arrays of numbers, so users can set /SwitchableOutput/output_1/Settings/Labels without hitting the "Invalid value types" error.

Closes #429

### DIFF
--- a/src/nodes/victron-virtual-functions.js
+++ b/src/nodes/victron-virtual-functions.js
@@ -1135,8 +1135,8 @@ export function validateVirtualDevicePayload (payload) {
     if (typeof value === 'number') return false
     if (typeof value === 'boolean') return false
     if (Array.isArray(value)) {
-      // Arrays are valid (for LightControls), but should contain only numbers
-      return !value.every(item => typeof item === 'number')
+      // Arrays are valid (for LightControls and Labels), but must be homogeneous (all numbers or all strings)
+      return !(value.every(item => typeof item === 'number') || value.every(item => typeof item === 'string'))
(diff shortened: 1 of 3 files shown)


src/nodes/victron-virtual-functions.js
Chris927, line 1131: `array of numbers` is now incorrect, should read `, array of numbers, or array of strings`.

This is what copilot is referring to, I think, but with a lot of words.

src/services/utils.js
Chris927, line 325: We should really get rid of the duplicated function with the same name in `src/nodes/victron-virtual-functions.js`, their content is already subtly different: The one handles `s2Signal`, the other doesn't.

I suggest to remove the function from `src/nodes/victron-virtual-functions.js`, and `require()` this one.

PR #433, opened 2026-03-25
Virtual energy meter, enable virtual devices to have dynamic properties

Chris927:
# Scope: Virtual energy meter

Implements a virtual energy meter, which can have different roles (grid meter, inverter, generator, etc).

# Scope: Virtual devices with dynamic properties

The virtual energy meter has a `Position` property, but not if it's a grid meter (the position is always on the "grid side"). To enable this requirement, the `properties` exported from a virtual device in `src/nodes/victron-virtual/device-type` can now be a function (not just an object), and we call that function with the node configuration. This way, the virtual device can have different properties, depending on the configuration.

### DIFF
--- a/src/nodes/victron-virtual.html
+++ b/src/nodes/victron-virtual.html
@@ -29,7 +29,7 @@
             grid_nrofphases: {value: 1, required: false},
             // energymeter
             energymeter_nrofphases: {value: 1, required: false},
-            energymeter_role: {value: "grid", required: false},
+            energymeter_role: {value: "gridmeter", required: false},
             // motordrive
             include_motor_temp: {value: false},
             include_controller_temp: {value: false},
--- a/src/nodes/victron-virtual/device-type/energymeter.js
+++ b/src/nodes/victron-virtual/device-type/energymeter.js
@@ -1,18 +1,32 @@
+const debug = require('debug')('victron-virtual')
+
+const ROLE_TO_SERVICE_TYPE = {
+  gridmeter: 'grid',
+  inverter: 'pvinverter',
+  generator: 'genset',
+  acload: 'acload',
+  evcharger: 'evcharger',
+  heatpump: 'heatpump'
+}
+
 const sharedProperties = {
-  'Ac/Current': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'A' : '', value: 0 },
   'Ac/Energy/Forward': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'kWh' : '', value: 0 },
   'Ac/Energy/Reverse': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'kWh' : '', value: 0 },
   'Ac/Power': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'W' : '' },
   'Ac/PowerFactor': { type: 'd', format: (v) => v != null ? v.toFixed(2) : '' },
-  'Ac/Voltage': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'V' : '' },
+  Connected: { type: 'i', format: (v) => v != null ? v : '', value: 1 },
   DeviceType: { type: 'i', format: (v) => v != null ? v : '', value: 0 },
   ErrorCode: { type: 'i', format: (v) => v != null ? v : '', value: 0 },
   IsGenericEnergyMeter: { type: 'i', format: (v) => v != null ? v : '', value: 1 },
   NrOfPhases: { type: 'i', format: (v) => v != null ? v : '', value: 1 }
 }
 
+function getServiceType (config) {
+  return ROLE_TO_SERVICE_TYPE[config.energymeter_role] || 'grid'
+}
+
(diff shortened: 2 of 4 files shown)


src/nodes/victron-virtual.html
```diff
@@ -27,6 +27,9 @@
             include_history_energy: {value: false},
             // grid
             grid_nrofphases: {value: 1, required: false},
+            // energymeter
+            energymeter_nrofphases: {value: 1, required: false},
+            energymeter_role: {value: "grid", required: false},
```
dirkjanfaber: The switch case in line 21 of src/nodes/victron-virtual/device-type/energymeter.js checks for `gridmeter`. And also in lie 362 below the value is set as `gridmeter`.

src/nodes/victron-virtual/device-type/energymeter.js
```diff
@@ -0,0 +1,66 @@
...
+  'Ac/PowerFactor': { type: 'd', format: (v) => v != null ? v.toFixed(2) : '' },
+  'Ac/Voltage': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'V' : '' },
+  DeviceType: { type: 'i', format: (v) => v != null ? v : '', value: 0 },
+  ErrorCode: { type: 'i', format: (v) => v != null ? v : '', value: 0 },
+  IsGenericEnergyMeter: { type: 'i', format: (v) => v != null ? v : '', value: 1 },
+  NrOfPhases: { type: 'i', format: (v) => v != null ? v : '', value: 1 }
+}
+
+function buildProperties (config) {
+  console.log('Building properties for energy meter with config:', config)
```
dirkjanfaber: Should be a `debug`

src/nodes/victron-virtual/index.js
dirkjanfaber, line 1: All other device types are explicitly registered in deviceModules. It works because of auto-discovery, but I don't want to rely on that.

src/nodes/victron-virtual/device-type/energymeter.js
```diff
@@ -0,0 +1,66 @@
+const sharedProperties = {
+  'Ac/Current': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'A' : '', value: 0 },
```
dirkjanfaber: `Ac/Current` and `Ac/Voltage` don't exist on real devices. Lets not add them for the virtual devices.

src/nodes/victron-virtual/device-type/energymeter.js
```diff
@@ -0,0 +1,66 @@
...
+  console.log('Building properties for energy meter with config:', config)
+
+  // make copy of sharedProperties to avoid modifying the original object
+  const properties = { ...sharedProperties }
+
+  switch (config.energymeter_role) {
+    case 'gridmeter':
+      break
+    default:
+      properties.Position = { type: 'i', format: (v) => v === 0 ? 'output' : 'input', value: 0 }
```
dirkjanfaber, line 38: Lets add:
```
properties.PositionIsAdjustable: { type: 'i', value: 0 }
```
Here too, to make it consistent with an actual device.

src/nodes/victron-virtual/device-type/energymeter.js
```diff
@@ -0,0 +1,66 @@
...
+  'Ac/Current': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'A' : '', value: 0 },
+  'Ac/Energy/Forward': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'kWh' : '', value: 0 },
+  'Ac/Energy/Reverse': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'kWh' : '', value: 0 },
+  'Ac/Power': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'W' : '' },
+  'Ac/PowerFactor': { type: 'd', format: (v) => v != null ? v.toFixed(2) : '' },
+  'Ac/Voltage': { type: 'd', format: (v) => v != null ? v.toFixed(2) + 'V' : '' },
+  DeviceType: { type: 'i', format: (v) => v != null ? v : '', value: 0 },
+  ErrorCode: { type: 'i', format: (v) => v != null ? v : '', value: 0 },
+  IsGenericEnergyMeter: { type: 'i', format: (v) => v != null ? v : '', value: 1 },
+  NrOfPhases: { type: 'i', format: (v) => v != null ? v : '', value: 1 }
```
dirkjanfaber, line 21: We also need:
```
Connected: { type: 'i', value: 1 }
```

src/nodes/victron-virtual/index.js
Chris927: What is "auto-discovery"? Do you mean looping over the files in `./device-type`? I think we should either loop over the files, or import/require everything explicitly, not both.

I suggest to leave it as is for this PR, but simplify afterwards.

Looping over the files makes sense if you expect anyone to use it as a feature, i.e. by placing files into the folder expecting them to be picked up on a restart. Otherwise we should not do it, and rather explicitly import/require.

PR #442, opened 2026-03-31
fix(virtual-switch): apply switch_1_show_ui_input config to ShowUIControl

dirkjanfaber:
The `ShowUIControl` iface property was always set to the hardcoded default value of `1` because the baseProperties loop never reads config values. Applied the same override pattern used for CustomName and Group.

So this is makes a6d7df22f8c7dcb8730866f05975169118b3e672 actually work.


Chris927:
When testing, the "Remote UI" option does not seem effective, see screenshots below. I was testing with betavrm. Is this still in development on the VRM side?

image


image

Chris927 (approved):
After testing, and after concluding that the sync issue observed is likely unrelated to this code change, I'm happy to approve.

PR #445, opened 2026-04-01
fix(virtual): use httpAdminRoot prefix for virtual-device-types endpoint

### LINKED ISSUE #440: No virtual device types listed
**Describe the bug**
When running stand alone in a node-red that is not running on the Cerbo, I not longer see any and device types listed in the popup.

dirkjanfaber:
Hardcoded /victron/virtual-device-types broke when Node-RED was deployed under a subpath. Now constructs the URL the same way victron-nodes.html does, using RED.settings.httpNodeRoot || RED.settings.httpAdminRoot.

Also documents the correct pattern in NODE_EDIT_DIALOG_DESIGN_GUIDE.md to prevent recurrence.

Closes #440, Closes #443


Chris927 (approved):
Looks good, tested with a local `httpNodeRoot`, working fine.

PR #453, opened 2026-04-13
feat(virtual-indicator): add virtual indicator node

dirkjanfaber:
Adds a new Virtual Indicator node that creates a GenericInput device on D-Bus, exposing indicator values on the GX device display and VRM remote console.

Indicator types:
- Discrete: labeled state (e.g. Off/On), configurable comma-separated labels with support for reserved GX keywords (/off, /on, /open, /closed, etc.)
- Value: numeric value with optional unit and decimal precision
- Value with range: numeric value with configurable min/max range indicator
- Temperature: temperature value with range indicator (no unit/label/decimals)

Node features:
- Accepts { Value, Status } or full D-Bus path keys as msg.payload
- Persists value and status across Node-RED restarts
- Show in UI control for local GX display and/or VRM remote console
- Node status shows type and 'no value' when uninitialised, value+unit once set
- Custom name and group are required fields (red border when empty, blocks deploy)
- Unit, primary label and decimals fields hidden for Temperature type
- Edit panel follows virtual switch style: no decorative label icons, tooltip icons only, with overwrite warning on custom name and group tooltips


src/nodes/victron-virtual-indicator.html
```diff
@@ -0,0 +1,214 @@
...
+                const isDiscrete = type === 0
+                const isTemperature = type === 3
+
+                $('#indicator-range-rows').toggle(hasRange)
+                $('#indicator-discrete-rows').toggle(isDiscrete)
+                $('#indicator-unit-row').toggle(!isDiscrete && !isTemperature)
+                $('#indicator-primary-label-row').toggle(!isDiscrete && !isTemperature)
+                $('#indicator-decimals-row').toggle(!isDiscrete && !isTemperature)
+
+                if (typeof renderIndicatorDocBox === 'function') {
```
Chris927: Do we need this condition? The function should always be there, unless we have a bug, IMU.

(Same question for the `if` statements below.)

src/nodes/victron-virtual-indicator.js
```diff
@@ -0,0 +1,396 @@
...
+module.exports = function (RED) {
+  let hasRunOnce = false
+  let globalTimeoutHandle = null
+  const nodeInstances = new Set()
+
+  function VictronVirtualIndicatorNode (config) {
+    RED.nodes.createNode(this, config)
+    const node = this
+
+    const address = process.env.NODE_RED_DBUS_ADDRESS
```
Chris927: The logic to determine `address`, and also logic related to `pendingCallsToSetValuesLocally` is duplicated: It's present here and in `victron-virtual-switch.js`, `victron-virtual/index.js`. If we don't refactor this to be DRY, we will likely break it some point in the future when making changes, but not in all those places identically.

Suggestion: We define a `function VictronVirtualCommonBaseNode(config) { ... }`, which contains such logic in a single place.

src/nodes/victron-virtual-indicator.js
```diff
@@ -0,0 +1,396 @@
...
+    const debouncedSetters = new Map()
+
+    function shouldApplyImmediately (key) {
+      if (node.ifaceDesc && node.ifaceDesc.properties && node.ifaceDesc.properties[key]) {
+        return node.ifaceDesc.properties[key].immediate === true
+      }
+      return false
+    }
+
+    function getDebouncedSetter (key) {
```
Chris927: Debouncing logic is also duplicated, needs to be refactored.

src/nodes/victron-virtual-indicator.js
```diff
@@ -0,0 +1,396 @@
...
+    this.on('input', function (msg, _send, done) {
+      if (!node.setValuesLocally) {
+        node.pendingCallsToSetValuesLocally.push([msg, done])
+        debugInput(`Node ${node.id} not ready, queuing message`)
+        return
+      }
+      handleInput(msg, done)
+    })
+
+    function instantiateDbus (self) {
```
Chris927, line 111: Also duplicated here and in `victron-virtual-switch.js` and `victron-virtual/index.js`.

src/services/virtual-indicator.js
```diff
@@ -0,0 +1,124 @@
...
+    { name: 'Settings/Group', type: 's', value: config.group || '', persist: false },
+    {
+      name: 'Settings/ShowUIInput',
+      type: 'i',
+      value: config.show_ui_input ?? 1,
+      persist: true,
+      format: (v) => {
+        if (v === 0) return 'Hidden'
+        const parts = []
+        if (v & 0b001) parts.push('All UIs')
```
Chris927: Why is there an `All UIs` option? As it is a bitfield, the "all" option can be expressed by all bits being set.

I suggest to not have an `All UIs` option, as the current implementation is ambiguous, `0b001` means the same as `0b110`.

I just saw that these values are defined [here](https://github.com/victronenergy/venus/wiki/dbus#generic-input-api); if this is really what is wanted, then there should be an explanation of the difference between "All UIs" and `0b110`.

src/nodes/victron-virtual-indicator.js
```diff
@@ -0,0 +1,396 @@
...
+        self.bus = process.env.DBUS_SESSION_BUS_ADDRESS ? dbus.sessionBus() : dbus.systemBus()
+      }
+
+      if (!self.bus) {
+        node.warn('Could not connect to the DBus session bus.')
+        node.status({ color: 'red', shape: 'dot', text: 'Could not connect to the DBus session bus.' })
+        return
+      }
+
+      const serviceName = `com.victronenergy.switch.virtual_${self.id}`
```
Chris927: Is it intentional to have a virtual indicator show up as switch?

Chris927:
When testing via betavrm, I only got this:

image

Chris927 (dismissed):
I struggled to test (see comment about betavrm), and I am concerned about lots of code duplication, see comments for details.

dirkjanfaber:
> I struggled to test (see comment about betavrm), and I am concerned about lots of code duplication, see comments for details.

This isn't on VRM (yet). Only way to test it is via gui-v2 (so via https://venus.local/gui-v2). And run one of the latest beta firmwares of Venus.

Chris927:
> > I struggled to test (see comment about betavrm), and I am concerned about lots of code duplication, see comments for details.
> 
> This isn't on VRM (yet). Only way to test it is via gui-v2 (so via https://venus.local/gui-v2). And run one of the latest beta firmwares of Venus.

Ah, thanks! I will test there later today.

Chris927:
Here an odd test result:

This indicator

image

... looks correct on gui-v2

image

... but incorrect in the Node-RED flow:

image

The reason is likely that I copied the node from another indicator, which had "kWh" as unit, but then I changed the indicator type to "Discrete" (which does not have a unit in its configuration).

dirkjanfaber:
Resolved all open issues:

_Code duplication (#2, #3, #4)_

Extracted into three standalone functions in `victron-virtual-dbus-helpers.js` (registerInputHandler, flushPendingInputs, createDebouncedSetters) rather than a base class. All three node files now call these. A shared base class was avoided because each node calls RED.nodes.createNode directly and has a different proceed()
 setup - the differences outweigh the commonality.

_Switch service type (#6)_

Intentional. The GenericInput API is defined under `com.victronenergy.switch` in the Venus OS D-Bus spec; there is no dedicated indicator type. The `vindic_` prefix distinguishes indicators from virtual switches on the bus.

_GenericInput/0/Name_

Now set to the indicator type name (Discrete, Value, etc.) and marked read-only. The custom name is still exposed via CustomName.

_ShowUIInput_

Defaults to `1` (both local and remote UI). The UI renders two separate checkboxes mapping to the bitfield values from the spec - values `0`, `1`, `2`, `4` only. No more ambiguous "All UIs" option.

_HTML defensive check (#1)_

Removed that.

src/nodes/victron-virtual-indicator.js
```diff
@@ -0,0 +1,323 @@
...
+const debugInput = require('debug')('victron-virtual-indicator:input')
+const debugConnection = require('debug')('victron-virtual-indicator:connection')
+const { DEBOUNCE_DELAY_MS } = require('./victron-virtual-constants')
+const { validateVirtualDevicePayload, debounce } = require('../services/utils')
+const { createIndicatorProperties, updateIndicatorStatus, INDICATOR_INPUT_KEY } = require('../services/virtual-indicator')
+const { filterInactiveVirtualDevices } = require('../services/virtual-device-cleanup')
+const { getTcpBusAddress, callAddSettingsWithRetry, getDeviceInstance, registerInputHandler, flushPendingInputs, createDebouncedSetters } = require('./victron-virtual-dbus-helpers')
+
+process.on('unhandledRejection', (reason, promise) => {
+  console.error('Unhandled rejection in victron-virtual-indicator:', reason)
```
Chris927: We should handle unhandled rejections only once (for the whole nodejs process), as handlers attached to the `process` object have process scope, not file scope.

I noticed that we already have other `unhandledRejection` event handlers elsewhere. I suggest we remove this one, and add an issue to "clean up unhandledRejection handlers".

Chris927 (changes requested):

src/services/virtual-indicator.js
```diff
@@ -0,0 +1,136 @@
...
+    },
+    { name: 'Status', type: 'i', value: 0, persist: true, format: (v) => INDICATOR_STATUS_NAMES[v] ?? String(v) },
+    { name: 'Name', type: 's', value: INDICATOR_TYPE_NAMES[indicatorType] || 'Indicator', readonly: true },
+    { name: 'Settings/CustomName', type: 's', value: config.customname || '', persist: false },
+    { name: 'Settings/Group', type: 's', value: config.group || '', persist: false },
+    {
+      name: 'Settings/ShowUIInput',
+      type: 'i',
+      value: config.show_ui_input ?? 1,
+      persist: true,
```
Chris927: Making this persistent results in configuration changes to this property becoming ineffective: We [load persisted state](https://github.com/victronenergy/node-red-contrib-victron/blob/d0cd8d2f11e0cb577eb52858149e8d541a945a4e/src/nodes/victron-virtual-indicator.js#L172) *after* this code runs, so the persisted value always overwrites.

A possible solution: We can query the cache, similar to how [we do it for switch name and group](https://github.com/victronenergy/node-red-contrib-victron/blob/fff2bb689f0398a5b0844c38a4fc793243b381fc/src/nodes/victron-virtual-functions.js#L891-L909). If we don't do this, then the visibility can be changed via dbus only, not via the node configuration in Node-RED.

Chris927: Making this persistent results in configuration changes to this property becoming ineffective: We [load persisted state](https://github.com/victronenergy/node-red-contrib-victron/blob/d0cd8d2f11e0cb577eb52858149e8d541a945a4e/src/nodes/victron-virtual-indicator.js#L172) *after* this code runs, so the persisted value always overwrites.

A possible solution: We can query the cache, similar to how [we do it for switch name and group](https://github.com/victronenergy/node-red-contrib-victron/blob/fff2bb689f0398a5b0844c38a4fc793243b381fc/src/nodes/victron-virtual-functions.js#L891-L909). If we don't do this, then the visibility can be changed via dbus only, not via the node configuration in Node-RED.

Chris927: Making this persistent results in configuration changes to this property becoming ineffective: We [load persisted state](https://github.com/victronenergy/node-red-contrib-victron/blob/d0cd8d2f11e0cb577eb52858149e8d541a945a4e/src/nodes/victron-virtual-indicator.js#L172) *after* this code runs, so the persisted value always overwrites.

A possible solution: We can query the cache, similar to how [we do it for switch name and group](https://github.com/victronenergy/node-red-contrib-victron/blob/fff2bb689f0398a5b0844c38a4fc793243b381fc/src/nodes/victron-virtual-functions.js#L891-L909). If we don't do this, then the visibility can be changed via dbus only, not via the node configuration in Node-RED.

Chris927: Making this persistent results in configuration changes to this property becoming ineffective: We [load persisted state](https://github.com/victronenergy/node-red-contrib-victron/blob/d0cd8d2f11e0cb577eb52858149e8d541a945a4e/src/nodes/victron-virtual-indicator.js#L172) *after* this code runs, so the persisted value always overwrites.

A possible solution: We can query the cache, similar to how [we do it for switch name and group](https://github.com/victronenergy/node-red-contrib-victron/blob/fff2bb689f0398a5b0844c38a4fc793243b381fc/src/nodes/victron-virtual-functions.js#L891-L909). If we don't do this, then the visibility can be changed via dbus only, not via the node configuration in Node-RED.

src/services/virtual-indicator.js
```diff
@@ -0,0 +1,136 @@
...
+      ...(rangeMax !== undefined ? { max: rangeMax } : {})
+    },
+    { name: 'Status', type: 'i', value: 0, persist: true, format: (v) => INDICATOR_STATUS_NAMES[v] ?? String(v) },
+    { name: 'Name', type: 's', value: INDICATOR_TYPE_NAMES[indicatorType] || 'Indicator', readonly: true },
+    { name: 'Settings/CustomName', type: 's', value: config.customname || '', persist: false },
+    { name: 'Settings/Group', type: 's', value: config.group || '', persist: false },
+    {
+      name: 'Settings/ShowUIInput',
+      type: 'i',
+      value: config.show_ui_input ?? 1,
```
Chris927: `config.show_ui_input` is initialized with a `Number`, but changes to a `String`, if I change the corresponding UI control before deploying a newly created indicator. As a result, the value persisted will be a `String` instead of a `Number`.

Solution:

```patch
diff --git a/src/services/virtual-indicator.js b/src/services/virtual-indicator.js
index f8fd35a..ae6fb11 100644
--- a/src/services/virtual-indicator.js
+++ b/src/services/virtual-indicator.js
@@ -45,7 +45,7 @@ function createIndicatorProperties (config, ifaceDesc, iface) {
     {
       name: 'Settings/ShowUIInput',
       type: 'i',
-      value: config.show_ui_input ?? 1,
+      value: Number(config.show_ui_input ?? 1),
       persist: true,
       format: (v) => ({ 0: 'Hidden', 1: 'All UIs', 2: 'GX device', 4: 'VRM' }[v] ?? String(v))
     },
```

src/services/virtual-indicator.js
dirkjanfaber: I believe it is safe to remove the persistency for this. Making the edit panels config leading.

src/nodes/victron-virtual-indicator.js
dirkjanfaber: https://github.com/victronenergy/node-red-contrib-victron/issues/469

Chris927 (approved):

PR #454, opened 2026-04-14
Feat/s2 node

dirkjanfaber:
In order to work correctly together with the (under development) s2 resource manager node, a few updates on the virtual acload where needed. 
Some outdated paths where removed and s2 power measurement support was added.


Chris927 (approved):
Was not able to test yet, but I believe this is low risk, as it only activates if S2 support is enabled. And S2 support is currently in development anyway.

PR #457, opened 2026-04-15
feat(virtual): auto-update LastEvContact on any EV property change

dirkjanfaber:
Adds o`nPropertyChanged` to the ev device module so that any incoming property update automatically sets `LastEvContact` to the current Unix timestamp. Uses the `skipOnPropertyChanged` guard (same pattern as #403) to prevent recursive emitCallback calls.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,8 @@
       "version": "1.7.8",
       "license": "MIT",
       "dependencies": {
-        "dbus-victron-virtual": "^0.1.33",
+        "dbus-native-victron": "^0.4.5",
+        "dbus-victron-virtual": "^0.1.34",
         "debug": "^4.4.0",
         "lodash": "^4.17.21",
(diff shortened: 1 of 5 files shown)


Chris927 (changes requested):
[Here](https://github.com/Chris927/node-red-contrib-victron/tree/virt-ev-lastevcontacted-alternative) is an alternative implementation, which uses the approach suggested on #403: This approach is cleaner, will support future (more complex) "derived properties" better, but requires changes in `dbus-victron-virtual`: These changes are in [this feature branch](https://github.com/Chris927/dbus-victron-virtual/tree/on-properties-changed); once that is reviewed (by @dirkjanfaber ;)), and if we are generally happy with this approach, I can cut a release there, and the current dependency to a commit-hash can be replaces with a dependency onto a new version of `dbus-victron-virtual`.

Chris927 (approved):

PR #459, opened 2026-04-20
fix(virtual): restore ProductId for generator and e-drive device types

### LINKED ISSUE #458: Virtual generator: `/ProductId` missing
Venus OS 3.80~14.

The virtual ac- and dc genset services are missing the `/ProductId` path. This prevents it from showing up in the UI properly:

Image

Image

dirkjanfaber:
Fixes #458.

`ifaceDesc.productType` was set to `config.device` (`'generator'` or `'e-drive'`), neither of which exists in `dbus-victron-virtual`'s products list. This caused `addVictronInterfaces` to silently skip `ProductId`, `ProductName`, and `Mgmt/*` - preventing generators from appearing in the Venus OS UI.

Without the override, the library correctly extracts `'genset'`, `'dcgenset'`, or `'motordrive'` from the service name.


src/nodes/victron-virtual/index.js
```diff
@@ -515,10 +515,6 @@ module.exports = function (RED) {
           }
         }
 
-        if (dbusServiceType !== config.device) {
-          ifaceDesc.productType = config.device
```
Chris927, line 490: By removing this, we change (break?) 970cf0c0. Specifically, the product ID for a virtual  energy meter changes from 0xc06f ("energy meter") to 0xc06a ("AC load"), which is not what we want.

Chris927: I'm not sure what the right fix is, though: We should definitely add unit tests to cover 970cf0c to avoid accidentally breaking it. Once done, how about removing `        if (dbusServiceType !== config.device) {
`, but then adding specific logic to the specific `getServiceType()` functions under `src/nodes/victron-virtual/device-type/*.js`?

Chris927 (dismissed):
Breaks an earlier fix, see comments on what I *think* we should do.

src/nodes/victron-virtual/index.js
dirkjanfaber: Good catch. We indeed don't want to break that. The approach of first adding the unit test is an excellent idea. Also not sure what the final solution is going to be, but at least we should be able to prevent breaking things.
I'll work on the test + a fix, starting at your suggestion.

Chris927 (approved):
Looks good, fixes the issue described, and adds some nice tests.

PR #463, opened 2026-04-30
fix(virtual): resolve code findings from PR #386 review

### LINKED ISSUE #400: Code findings after PR 386
The review process of PR #386 found a couple of issues that need resolving (and postponed for this issue as the findings where already in the code before migrating). 

All can be found in the `src/nodes/victron-virtual/index.js` file:
- [ ] Why `stringify()` and then `parse()`. Line `145` (https://github.com/victronenergy/node-red-contrib-victron/pull/386#discussion_r2882523899)
- [ ] Dubious usage of `actualDev` vs `dev`: We use actualDev in the next line, but further down we use dev. Lines `108` (https://github.com/victronenergy/node-red-contrib-victron/pull/386#discussion_r2882538101) and `131` (https://github.com/victronenergy/node-red-contrib-victron/pull/386#discussion_r2882553553)
- [ ] This object definition duplicates the object we return within the previous `if()`. I suggest to refactor this, so that the for (const key in ...) below iterates zero times if there are no properties. The return value at the end of the function will be the same (as the object returned in the previous if()). Around line `139` (https://github.com/victronenergy/node-red-contrib-victron/pull/386#discussion_r2882551947)

And add extra test to increase code coverage:
- [ ] https://github.com/victronenergy/node-red-contrib-victron/pull/386#discussion_r2882493793

Added later: 
- [ ] Fix the mentioned duplication from #430

dirkjanfaber:
## Summary

- Removes the dead `validateVirtualDevicePayload` export from `victron-virtual-functions.js` - it was never imported in `victron-virtual-browser.js` and not exposed via `window.__victron`, making it unreachable in the browser bundle. The canonical version lives in `src/services/utils.js`. This closes the duplication flagged when merging #430.
- Simplifies `getIface` in `victron-virtual/index.js`: creates the `result` object once and returns early, eliminating the duplicate `{ emit: function() {} }` literal; also drops the unnecessary `JSON.parse(JSON.stringify(...))` since only `.value` and `.type` are read from each property, never mutated.
- Adds a format function null-safety contract test to `victron-virtual-device-types.test.js` to verify that all device type `format()` functions handle `null` input without throwing (coverage improvement requested in the PR #386 review).

The `actualDev` vs `dev` naming issue from the checklist was already resolved in a prior refactoring.

No conflict with open PR #453: that PR's additions to `victron-virtual-functions.js` are in the middle of the file, its new `victron-virtual-indicator.js` imports `validateVirtualDevicePayload` from `utils.js` (not the browser file), and its `index.js` changes only touch debounce/pending-call logic.

Closes #400


Chris927 (approved):
Looks good. I did a brief smoke test.

PR #466, opened 2026-05-04
fix(system): filter null D-Bus values from input/output node dropdowns

dirkjanfaber:
Paths whose cached D-Bus value is null (the empty variant []) are not applicable to the current device configuration and should not appear in the measurement dropdown of regular input/output nodes.

Custom nodes (getCachedServices) are intentionally unaffected and continue to expose null-valued paths.

Fixes a bug reported by @warwickchapman

### DIFF
--- a/src/services/victron-system.js
+++ b/src/services/victron-system.js
@@ -59,7 +59,8 @@ class SystemConfiguration {
               )) &&
               (expandedPathObj.mode === 'both' ||
               (isOutput && expandedPathObj.mode === 'output') ||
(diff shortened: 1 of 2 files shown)


Chris927:
As discussed, as the value may not always be `null`, it'll be better to not filter, but rather display `(null)` in the dropdown.

@dirkjanfaber will update the PR accordingly.

Chris927 (approved):
Great to see that you could also simplify some complex boolean logic :)

PR #468, opened 2026-05-04
feat(virtual-switch): show customname, group and type in node label

### LINKED ISSUE #464: [Feature] Virtual Switch: show configuration name if name is empty
**Is your feature request related to a problem? Please describe.**

When a virtual switch is added and the name is left empty, then it shows `Virtual Switch` as the node name.

**Describe the solution you'd like**
If the name field is empty, it should use the configuration name. This is default for all non Victron nodes.

Image

Image

Image

Image

dirkjanfaber:
When the Name field is left empty, the virtual switch and indicator nodes now fall back to a label built from the configured custom name, group and switch/indicator type (e.g. "Fan (Ventilation) [Toggle]") instead of a plain "Virtual Switch" / "Virtual Indicator" string.

Shared getVirtualNodeLabel() added to victron-virtual-functions.js and exposed on window.__victron so both nodes use the same logic. INDICATOR_TYPE_LABELS also added to window.__victron for indicator type name lookup.

Closes #464

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -1459,11 +1459,26 @@
     $('#node-input-outputs').val(outputs);
   }
 
+  /**
+   * Returns the Node-RED palette label for a virtual node.
+   * Priority: name -> customname + group + typeName -> fallback + typeName
+   * @param {{ name?: string, customname?: string, group?: string, typeName?: string, fallback?: string }} opts
+   * @returns {string}
+   */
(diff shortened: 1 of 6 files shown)


Chris927:
Explanation on my force-push: I used the web-merge tool of github to resolve conflicts, but somehow the conflict resolution did not work. I then fixed it via local CLI and force-pushed.

Chris927 (approved):
Tested and working, code is clean.

Apologies for messing up the conflict resolution resulting in some unnecessary commits 🙈. I assume you will merge with a single squash-commit so that we have just a single commit in the main branch.

PR #475, opened 2026-05-07
feat(virtual): shortcut control and passthrough mode for virtual switch / indicator

dirkjanfaber:
Two UX improvements for the virtual switch and indicator nodes.

-   The doc box inside each switch and indicator edit panel now marks the most relevant D-Bus path with a ⚡ bolt icon. Sending a plain value as `msg.payload` (instead of a `{"path": value}` object) targets that path directly: no wrapping needed for the common case.
  The bolt replaces the bullet for that list item and shows a tooltip on hover explaining the shortcut.                                                        
  
- Three-state switch: passthrough mode
   Adds an Apply input dropdown to the three-state switch configuration:
  - _Always_: Incoming values are always written to D-Bus (current behaviour).
  - _Auto only_: Values are only applied when the switch is in Auto mode (Auto=1) 
  With Auto only (the default for new nodes), a user can put the switch into manual mode on the GX device to take direct control, and Node-RED automations into the switch are silently ignored until auto mode is restored.
                                                                                                                                                               
  Existing nodes without the field saved behave as _Always_, so there is no behaviour change for deployed flows.

And also included an example flow for the three-state switch wired from automation inject nodes through to a virtual indicator, demonstrating the passthrough mode.


examples/Switch - Three-state automation control.json
Chris927, line 1: The filename should end in `.json`, and a "/" in the filename may cause issues on some systems.

src/services/virtual-switch.js
```diff
@@ -643,11 +644,56 @@ function emitInitialSwitchOutputs (config, node) {
...
+  return { [defaultPath]: payload }
+}
+
+/**
+ * Returns true if the incoming payload should be written to D-Bus given the current Auto state.
+ * For non-three-state switches, always returns true. For three-state switches, the
+ * switch_1_passthrough_mode config field controls when writes are allowed:
+ *   'always'      - always write (same as other switch types)
+ *   'auto_only'   - only write when Auto=1 (default; prevents automations from overriding manual control)
+ *   'manual_only' - only write when Auto=0
```
Chris927: Really? `manual_only` is neither in the logic below, nor in the config form in the frontend. Maybe a documentation error?

src/nodes/victron-virtual-switch.js
```diff
@@ -94,6 +104,12 @@ module.exports = function (RED) {
         }
       }
 
+      if (!shouldApplyPayloadToDBus(config, node.iface)) {
```
Chris927: We disallow switching to auto, if msg.payload['SwitchableOutput/output_1/Auto'] === 1, so we cannot switch to manual mode, e.g. via the demo flow.

Solution is probably to pass the payload into this function call, then allow the payload, if `.../Auto === 1` is in the payload.

Chris927 (changes requested):
As discussed.

src/nodes/victron-virtual-indicator.js
```diff
@@ -26,30 +26,17 @@ module.exports = function (RED) {
...
-        }
-      }
-      return expanded
-    }
-
     function handleInput (msg, done) {
       node.send(msg)
 
-      if (!msg || !msg.payload) {
+      if (!msg || msg.payload == null) {
```
Chris927: As I understand it, there might be a scenario where I want to set the default property to "null", and this condition would prevent this.

Probably this condition should be `!msg || msg.payload === undefined`?

src/nodes/victron-virtual-switch.js
```diff
@@ -55,7 +55,7 @@ module.exports = function (RED) {
       }
       node.send(outputs)
 
-      if (!msg || !msg.payload) {
+      if (!msg || msg.payload == null) {
```
Chris927: Same comment about `null` vs `undefined`.

src/nodes/victron-virtual-switch.js
```diff
@@ -55,7 +55,7 @@ module.exports = function (RED) {
...
         node.warn('Received message without payload. Expected: JavaScript object with at least one property/value.')
```
Chris927: This message should adjust to something like `... Expected: value for the default path, or a JavaScript object with keys/values`.

Chris927 (changes requested):
I tested manually, and it's working well.

Some small issues should still be fixed.

Chris927 (approved):

PR #477, opened 2026-05-11
EV updates

dirkjanfaber:
Two changes:
- feat(input): add victron-input-ev node for Electric Vehicle devices - We were missing that one. 
- feat(virtual): add LastUpdated/* timestamp paths to virtual EV device - Updated paths (see https://github.com/victronenergy/venus/issues/1633)


src/nodes/victron-virtual/device-type/ev.js
```diff
@@ -31,7 +31,34 @@ const properties = {
     value: 0,
     persist: 300
   },
-  LastEvContact: {
+  'LastUpdated/ChargingStarted': {
+    type: 'i',
+    format: (v) => {
```
Chris927: Looks like the format function is the same for all the `LastUpdated/...` properties, so it should be a single function, not copy-paste.

src/nodes/victron-virtual/device-type/ev.js
```diff
@@ -55,9 +82,22 @@ function initialize (config, _ifaceDesc, iface, _node) {
...
-    changes.LastEvContact = Math.floor(Date.now() / 1000)
+  const now = Math.floor(Date.now() / 1000)
+
+  if (!changes['LastUpdated/ProviderContact']) {
+    changes['LastUpdated/ProviderContact'] = now
+  }
+
+  const chargingFields = ['Soc', 'RangeToGo', 'Odometer', 'ChargingState']
+  if (chargingFields.some(f => f in changes) && !changes['LastUpdated/ChargingStarted']) {
+    changes['LastUpdated/ChargingStarted'] = now
```
Chris927: If charging started at time t, and some logic updates any of those fields at t2 > t, then the *actual* time when charging started will be lost. Is this intentional?

Chris927 (changes requested):

Chris927 (approved):

PR #479, opened 2026-05-11
fix(virtual): ensure pulsemeter UI controls only show for pulsemeter

Chris927:


dirkjanfaber (approved):
Yes, really needed fix and failed to notice this before.

Issue #418, opened 2026-03-17
[Feature] add hysterisis to the conditional input


dirkjanfaber:
While the conditional inputs work great, it is not making it easy to do something like: switch the device on when the SOC is above 80% and switch off again when it drops below 70%.

One of the challenges is, is to implement this is such a way that it doesn't directly overwhelm the user that just wants to switch the device when the SOC is above 80%. So it probably should go under an "advanced control" section or so.

dirkjanfaber:
Closing as 11aaf47 made it into master.

PR #481, opened 2026-05-16
feat(conditional): add hysteresis to conditional input node (#418)

dirkjanfaber:
Adds hysteresis support to prevent rapid switching near a threshold. When enabled, the condition stays TRUE until the value crosses a separate turn-off threshold (e.g. turn ON above 80%, turn OFF below 70%).

- Backend: evaluateConditionWithHysteresis uses inverse operator against the turn-off threshold when the previous result was TRUE
- UI: collapsible Advanced section under Condition 1 with enable checkbox, turn-off threshold input, and live validation
- Validation: warns when threshold is empty, on wrong side, or operator is == / != (unsupported)
- CSS: fix info/warning/tip boxes overflowing victron-condition-section; reduce condition row input width by 10px to add clearance from border

### DIFF
--- a/src/nodes/victron-nodes.html
+++ b/src/nodes/victron-nodes.html
@@ -130,6 +130,37 @@
                     <label for="node-input-condition1-threshold">Threshold</label>
                     <input type="number" id="node-input-condition1-threshold" placeholder="e.g., 50" step="any" autocomplete="off" data-lpignore="true">
                 </div>
+
+                <!-- Hysteresis (Advanced) -->
+                <div class="form-row victron-condition-row">
+                    <label>&nbsp;</label>
+                    <button type="button" id="toggle-hysteresis-btn" class="victron-add-condition-btn">
+                        <i class="fa fa-chevron-right" id="hysteresis-chevron"></i> Advanced: Hysteresis
+                    </button>
+                </div>
+                <div id="hysteresis-panel" class="hidden">
+                    <div class="form-tips">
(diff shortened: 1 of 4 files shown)


src/nodes/victron-nodes.html
```diff
@@ -422,13 +453,50 @@
                 if (node.logicOperator) $('input[name="logicOperator"][value="' + node.logicOperator + '"]').prop('checked', true);
             }
 
+            // Restore hysteresis state
+            if (node.condition1HysteresisEnabled) {
+                $('#hysteresis-panel').removeClass('hidden').show();
```
Chris927, line 458: [non-blocking] This could probably be simplified by using  `$(...).toggle(false)` and `$(...).slideToggle()`

src/nodes/victron-nodes.html
```diff
@@ -422,13 +453,50 @@
...
             // outputOnChange is now always enforced to true in the backend for conditional mode
             if (node.debounce !== undefined) $('#node-input-debounce').val(node.debounce);
         }
 
+        // Toggle hysteresis advanced panel
+        $('#toggle-hysteresis-btn').on('click', function() {
+            const $panel = $('#hysteresis-panel');
+            const $chevron = $('#hysteresis-chevron');
+            if ($panel.is(':visible')) {
+                $panel.slideUp(200);
```
Chris927, line 479: As per comment above: using `toggle()` etc should make magic number `200` unnecessary.

Chris927 (approved):
Looks good, but UI logic could be simplified; I consider this non-blocking, should refactor next time we touch this.

PR #482, opened 2026-05-18
fix(virtual): ev LastUpdated/Charging and Location only update on value change

dirkjanfaber:
- Only set LastUpdated/Charging when ChargingState actually differs from the current instance value; also fixes silent skip when ChargingState transitions to 0 (falsy)
- Only set LastUpdated/Location when a position field value actually differs from the current instance value
- Add test coverage for unchanged-value cases

### DIFF
--- a/src/nodes/victron-virtual/device-type/ev.js
+++ b/src/nodes/victron-virtual/device-type/ev.js
@@ -62,7 +62,7 @@ function onPropertiesChanged ({ changes, instance }) {
   }
 
   const positionFields = ['Position/Latitude', 'Position/Longitude', 'AtSite']
-  if (positionFields.some(f => f in changes && !changes['LastUpdated/Location']) {
+  if (positionFields.some(f => f in changes && !changes['LastUpdated/Location'])) {
     changes['LastUpdated/Location'] = now
   }
 
--- a/src/nodes/victron-virtual/device-type/ev.js
+++ b/src/nodes/victron-virtual/device-type/ev.js
@@ -62,7 +62,7 @@ function onPropertiesChanged ({ changes, instance }) {
   }
 
   const positionFields = ['Position/Latitude', 'Position/Longitude', 'AtSite']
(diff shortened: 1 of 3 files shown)


src/nodes/victron-virtual/device-type/ev.js
```diff
@@ -55,16 +54,20 @@ function initialize (config, _ifaceDesc, iface, _node) {
...
-  if (!changes['LastUpdated/ProviderContact']) {
-    changes['LastUpdated/ProviderContact'] = now
+  if (!changes['LastUpdated/EvContact']) {
+    changes['LastUpdated/EvContact'] = now
   }
 
   const positionFields = ['Position/Latitude', 'Position/Longitude', 'AtSite']
-  if (positionFields.some(f => f in changes) && !changes['LastUpdated/Position']) {
-    changes['LastUpdated/Position'] = now
+  if (positionFields.some(f => f in changes && changes[f] !== instance[f]) && !changes['LastUpdated/Location']) {
```
Chris927: I don't believe comparing the actual floating point values within the positionFields is useful to determine if the timestamp should be set or not: For a stationary EV, the location will likely change by some fraction all the time, even though the actual, physical location hasn't changed. The reason is that gps positions are inherently not precise, i.e. they may drift a few meters over time.

I believe it is semantically more robust to set the `LastUpdated` at value whenever there is a new value for a position field received, regardless if the value has changed, similar to how a unix "modified" timestamp on a file changes on every write, regardless if the write changes the file's content or not.

*If* you want to determine "has the EV moved", then some logic that looks at the deltas between consecutive positions (or positions within a time range) needs to be used, which is significantly more involved, and probably out of scope for this.

Chris927 (commented):
My only concern is how `LastUpdated/Location` is updated, see comment.

Let's discuss, then I'll re-review.

Chris927 (approved):

PR #487, opened 2026-05-25
Fix conditional debounce

### LINKED ISSUE #483: Output of conditional logic of Custom Input Node incorrect when using debouncing
**Describe the bug**

When using debounce on a Custom Input Node with conditional logic, the logic result is sometimes incorrect.

**To Reproduce**
1. Create a custom input node, e.g. reading `/SwitchableOutput/output_1/Dimming` of a virtual temperature switch, where the temperature switch has a range 10 to 50.
2. Set a rate limit of "10 msg/sec" (probably not relevant), set "debounce" to 2000, and enable conditional evaluation with a condition "> 25". Leave the rest of the configuration as per the defaults.
3. Deploy.
4. Change the temperature switch swiftly, crossing the threshold (of 25, in this case), and back.

**Expected behavior**

The value of the second output of the Custom Input Node should be correct, once 2 seconds have passed since the last change. But as per the video, this is not always the case.

**Screenshots**

The below video shows how the value 25.5 gets first represented as "true" (which is correct), then as "false" (which is incorrect).

https://github.com/user-attachments/assets/5e5b93a5-ba07-4b9d-b465-0b6fd020553c

**Flow**

Here is the node used in the example:

```json
[
  {
    "id": "7b59ddbd57cb569e",
    "type": "victron-input-custom",
    "z": "5c087902893e473b",
    "service": "com.victronenergy.switch/111",
    "path": "/SwitchableOutput/output_1/Dimming",
    "serviceObj": {
      "service": "com.victronenergy.switch/111",
      "name": "temp1 (111)"
    },
    "pathObj": {
      "path": "/SwitchableOutput/output_1/Dimming",
      "name": "/SwitchableOutput/output_1/Dimming",
      "type": "number",
      "value": 26
    },
    "name": "test-conditional",
    "onlyChanges": true,
    "roundValues": "1",
    "rateLimit": "10",
    "conditionalMode": true,
    "condition1Operator": ">",
    "condition1Threshold": 25,
    "condition1HysteresisEnabled": false,
    "condition1HysteresisThreshold": "",
    "condition2Enabled": false,
    "outputTrue": "true",
    "outputFalse": "false",
    "outputOnChange": false,
    "debounce": 2000,
    "x": 120,
    "y": 80,
    "wires": [
      [
        "23614d8ceff941f0"
      ],
      [
        "842b09abdb23f80b"
      ]
    ]
  }
]
```

 - Hardware: Cerbo GX, with Node-RED running locally

 - node-red-contrib-victron version: branch `feat/418-hysteresis-conditional-input` (commit b44d642)
 - Venus OS version: 3.80~14
 
**Additional context**
Even though observed on a feature branch, this is likely an issue with debouncing in the presence of conditional logic, unrelated to the feature branch.

Chris927:
The own implementation to debounce when setting the output of the conditional logic (in `src/nodes/victron-nodes.js`) behaved incorrectly in certain scenarios, see #483. By using debounce() from `src/services/utils.js` this is fixed.

Fixes #483


dirkjanfaber (approved):
Good find and a good fix.

PR #488, opened 2026-05-25
Fix multiple rejection handlers

### LINKED ISSUE #469: Remove redundant unhandledRejection defines (finding PR #453)
**Describe the bug**
Only one is needed:
```
process.on('unhandledRejection', (reason, promise) => {
  console.error('Unhandled rejection in victron-virtual-indicator:', reason)
})
``` 

**Additional context**
See #453

Chris927:
Adding an event handler for 'unhandledRejection' is global (for the process),
so it is unnecessary (and clutters the logs) to have multiple (in different
files).

I am leaving the handler in src/nodes/victron-virtual/index.js: This file
defines a node, and is definitely loaded unconditionally by Node-RED. This way
we can be sure that this handler is always registered.

Resolves #469


dirkjanfaber:
> I am leaving the handler in src/nodes/victron-virtual/index.js: This file
> defines a node, and is definitely loaded unconditionally by Node-RED. This way
> we can be sure that this handler is always registered.

I am tempted to believe that`src/services/victron-client.js`is more guaranteed to load than `src/nodes/victron-virtual/index.js` is.

Chris927:
> > I am leaving the handler in src/nodes/victron-virtual/index.js: This file
> > defines a node, and is definitely loaded unconditionally by Node-RED. This way
> > we can be sure that this handler is always registered.
> 
> I am tempted to believe that`src/services/victron-client.js`is more guaranteed to load than `src/nodes/victron-virtual/index.js` is.

I disagree: `src/services/victron-client.js` gets loaded indirectly (through our logic), but `src/nodes/victron-virtual/index.js` is (almost) directly loaded via the `node-red` section in `package.json`, through the entry `"victron-virtual"`:

```json
  "node-red": {
    "nodes": {
      "victron-client": "./src/nodes/config-client.js",
      "victron-event": "./src/nodes/victron-event.js",
      "victron-nodes": "./src/nodes/victron-nodes.js",
      "victron-virtual": "./src/nodes/victron-virtual.js",
      "victron-virtual-switch": "./src/nodes/victron-virtual-switch.js",
      "victron-virtual-indicator": "./src/nodes/victron-virtual-indicator.js",
      "victron-inject": "./src/nodes/victron-inject.js"
    },
    "version": ">=3.0.2"
  },
```

dirkjanfaber:
As discussed. Both get loaded, but the client one is more general, while the virtual one suggests that it is only used for the virtual nodes.

Chris927:
Updated as per discussion, please re-review.

dirkjanfaber (approved):
Approving.

Did have some thoughts on testability (e.g. adding a test that ensures there is exactly one entry point). But that is difficult as we'd need to define all possible entry points in that test too, which also needs manual maintenance. So not much added value there.

PR #489, opened 2026-05-28
fix(virtual-switch): allow leading-slash Auto key to exit manual mode

dirkjanfaber:
`shouldApplyPayloadToDBus` only checked `'SwitchableOutput/output_1/Auto'` (no leading slash), but users send d-bus paths with a leading slash. A payload of `{"/SwitchableOutput/output_1/Auto": 1}` was blocked with _"Ignored: switch in manual mode"_ even though its intent is to restore auto mode.

### DIFF
--- a/src/services/virtual-switch.js
+++ b/src/services/virtual-switch.js
@@ -682,7 +682,7 @@ function shouldApplyPayloadToDBus (config, iface, payload) {
   const mode = config.switch_1_passthrough_mode ?? 'always'
   if (mode === 'always') return true
 
-  if (payload && payload['SwitchableOutput/output_1/Auto'] === 1) return true
+  if (payload && (payload['SwitchableOutput/output_1/Auto'] === 1 || payload['/SwitchableOutput/output_1/Auto'] === 1)) return true
 
   const autoValue = iface ? iface['SwitchableOutput/output_1/Auto'] : null
   const isAuto = autoValue === 1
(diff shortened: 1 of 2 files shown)


warwickchapman:
Users send a leading slash because the in-node documentation includes a leading Slash:

image

src/services/virtual-switch.js
```diff
@@ -684,7 +682,7 @@ function shouldApplyPayloadToDBus (config, iface, payload) {
   const mode = config.switch_1_passthrough_mode ?? 'always'
   if (mode === 'always') return true
 
-  if (payload && payload['SwitchableOutput/output_1/Auto'] === 1) return true
+  if (payload && (payload['SwitchableOutput/output_1/Auto'] === 1 || payload['/SwitchableOutput/output_1/Auto'] === 1)) return true
```
Chris927, line 685: [not critical] We could avoid having (almost) the same string constant twice:

```js
const autoKey = "SwitchableOutput/output_1/Auto'";
if (payload && payload[autoKey] === 1 || payload['/' + autoKey] === 1)) {
  return true
}
```

... which makes it a bit more readable.

If we do this more often (i.e. allow an optional slash), we might want to consider a utils function:

```js
function matchesPayload({ payload, key, expected, isLeadingSlashOptional = false }) {
  ...
}
```

... but for now this would be overkill.

Chris927 (approved):
Approving it, but you might want to add my non-critical change for readability :)

PR #496, opened 2026-05-28
feat(indicator): add live SVG preview to virtual indicator edit dialog

### LINKED ISSUE #495: Please improve the GenericInput configuration pane
When configuring the virtual `GenericInput` instance, the example rendered below the input boxes would be better if it either displayed real values specified in the inputs (for CustomName and PrimaryLabel that is), or if it used "PrimaryLabel" and "CustomName" as placeholders rendered in the preview so that users can easily see what part of the component they are actually configuring... 

When custom name is not present the primary label becomes "State", when Unit is one of the predefined "/Speed", "/Volume", "/Temperature", the Primary Label becomes that string without a forward slash. This should probably be part of the tooltip...

Image

dirkjanfaber:
Replaces the static SVG images in the indicator doc box with a dynamic inline SVG that updates as the user types, addressing issue #495.

Per-type layout:
- Value/Discrete: simple card with label left, value right
- Value with range/Temperature: compact card with gradient bar and value right-aligned after the bar

Preview reflects:
- customname as the card title (falls back to input placeholder)
- primary_label as the card label (falls back to placeholder or reserved unit name for /Temperature, /Speed, /Volume)
- Reserved units resolved to display symbols (deg-C, km/h, L)
- Decimals field controls the placeholder value format
- Empty fields show placeholder text so layout is always meaningful

Tooltip updates: CustomName and Unit fields now document the GX device fallback behavior (State, keyword label override).

Disclaimer tooltip added to the doc box label.

Fixes #495

### DIFF
--- a/src/nodes/victron-virtual-functions.js
+++ b/src/nodes/victron-virtual-functions.js
@@ -9,6 +9,8 @@ import {
   SWITCH_THIRD_OUTPUT_LABEL
 } from './victron-virtual-constants'
 import { initializeTooltips } from './victron-common'
+/* global require */
+const escape = require('lodash/escape')
 
 // Re-export for browser/test use
 export {
@@ -1126,12 +1128,15 @@ export const INDICATOR_TYPE_LABELS = {
   3: 'Temperature'
 }
 
-function escapeHtml (str) {
-  return String(str)
-    .replace(/&/g, '&amp;')
-    .replace(/</g, '&lt;')
-    .replace(/>/g, '&gt;')
-    .replace(/"/g, '&quot;')
+export const INDICATOR_TYPE = {
+  DISCRETE: 0,
+  VALUE: 1,
+  VALUE_WITH_RANGE: 2,
+  TEMPERATURE: 3
+}
+
+function isTypeRange (type) {
+  return type === INDICATOR_TYPE.VALUE_WITH_RANGE || type === INDICATOR_TYPE.TEMPERATURE
 }
 
 export function renderIndicatorDocBox (type) {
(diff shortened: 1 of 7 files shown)


dirkjanfaber:
Note that we might also want to do such a thing to the virtual switches.

src/nodes/victron-virtual-functions.js
```diff
@@ -1126,6 +1126,14 @@ export const INDICATOR_TYPE_LABELS = {
   3: 'Temperature'
 }
 
+function escapeHtml (str) {
```
Chris927: Rather use `escape()` in the `lodash` package than implementing our own version.

src/nodes/victron-virtual-functions.js
```diff
@@ -1147,6 +1157,92 @@ export function renderIndicatorDocBox (type) {
...
+const RESERVED_UNIT_LABELS = { '/Temperature': 'Temperature', '/Speed': 'Speed', '/Volume': 'Volume' }
+const RESERVED_UNITS = { '/Temperature': '°C', '/Speed': 'km/h', '/Volume': 'L' }
+
+function buildSimpleCardSvg (title, cardLabel, valueText) {
+  const font = 'system-ui, -apple-system, sans-serif'
+  const safeTitle = escapeHtml(title)
+  const safeLabel = escapeHtml(cardLabel)
+  const safeValue = escapeHtml(valueText)
+  return (
+    '<svg width="368" height="88" viewBox="0 0 368 88" fill="none" xmlns="http://www.w3.org/2000/svg">' +
```
Chris927: Why concatenate so many strings? It is easier to read and more efficient to use multiline string with interpolation:

```js
return (`
<svg width=...
<text x= ... ${font} ...
`)

File: src/nodes/victron-virtual-functions.js [hunk 3]
```diff
@@ -1147,6 +1157,92 @@ export function renderIndicatorDocBox (type) {
...
+    '</svg>'
+  )
+}
+
+export function updateIndicatorLivePreview () {
+  const $preview = $('#indicator-live-preview')
+  if (!$preview.length) return
+
+  const type = parseInt($('#node-input-indicator_type').val(), 10)
+  const isRange = type === 2 || type === 3
```
Chris927: Magic numbers make this hard to read. Rather define the types in some place, e.g. `const INDICATOR_TYPE_TEMPERATURE = 13` (or whatever), then this logic becomes `type === INDICATOR_TYPE_POWER || type === INDICATOR_TYPE_TEMPERATURE`.

If such logic is required more often, define a function `isTypeRange(type)`, then this becomes `const isRange = isTypeRange(type)`, or you don't define the variable but use the function whenever you need to know it's a range.

Chris927 (changes requested):
Super-useful to have such a live preview!

I've got a few changes I suggest.

src/nodes/victron-virtual-functions.js
```diff
@@ -9,6 +9,8 @@ import {
   SWITCH_THIRD_OUTPUT_LABEL
 } from './victron-virtual-constants'
 import { initializeTooltips } from './victron-common'
+/* global require */
+const escape = require('lodash/escape')
```
Chris927: Using `require()` breaks in the browser ("require undefined" or similar).

What works in my understanding:

```js
import escape from 'lodash/escape'
```

... which is resolved by rollup, so that in the bundled version (`resources/victron-virtual-functions.js`) there is no "import" statement any more (and also no "require()"), so it works in the browser.

Chris927 (changes requested):
The "require()" does not work as is, as per manual testing, otherwise it looks good.

Chris927 (approved):

PR #498, opened 2026-05-29
fix: restore sync node registration to prevent grey nodes on startup

### LINKED ISSUE #494: Grey nodes on startup (fixed after refresh)
**Describe the bug**
Hard to trace bug, but occasionally the Victron nodes start grey instead of blue .

Image

**To Reproduce**
* unclear, happens mostly on startup

**Additional context**

Seems to be fixed after a refresh. Cause might be traced to the order in which the styles get loaded. The above screenshot shows an inspect of such a node. Searching for `rgb(221, 221, 221);` could be a starting point to find the actual cause.

dirkjanfaber:
Node-RED's appendConfig() waits for all `<script src>` tags to load before executing a module's inline scripts. Since `registerType` calls live in those inline scripts, adding `<script src="victron-common.js">` (and `victron-virtual-functions.js`) to the HTML file headers made node type registration async - introducing a race where `loadFlows()` could render the canvas before registration completed, leaving nodes grey.

Replace the header `<script src>` tags with a guard-based dynamic injection at the END of each file's inline script block, after all `registerType` calls. Node-RED's loader sees no deferred scripts and runs registration synchronously (as it did before 1.7.0). The external scripts still load in the background, well before any user can open an edit dialog.

The virtual device files (`victron-virtual`, `victron-virtual-switch`, `victron-virtual-indicator`) do not lazy-load `victron-common.js` because their `oneditprepare` functions only use `window.__victron.*` from `victron-virtual-functions.js`, which already bundles the same functions. The non-virtual files (`victron-nodes`, `victron-inject`, `victron-event`) do lazy-load it as they use `victronCommon` directly.

Regression introduced in 829026d. Fixes #494.

### DIFF
--- a/src/nodes/victron-event.html
+++ b/src/nodes/victron-event.html
@@ -1,5 +1,4 @@
 <link rel="stylesheet" href="resources/@victronenergy/node-red-contrib-victron/victron-styles.css">
-<script src="resources/@victronenergy/node-red-contrib-victron/victron-common.js"></script>
 
 <script type="text/javascript">
     RED.nodes.registerType('victron-event', {
@@ -27,6 +26,13 @@
             }
         }
     });
+
+    if (!document.querySelector('script[data-victron-common]')) {
+        var _scr = document.createElement('script');
+        _scr.setAttribute('data-victron-common', '');
+        _scr.src = 'resources/@victronenergy/node-red-contrib-victron/victron-common.js';
+        document.head.appendChild(_scr);
+    }
 </script>
 
 <script type="text/html" data-template-name="victron-event">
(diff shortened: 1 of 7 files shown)


dirkjanfaber:
A note on testing, as that isn't perse easy to do. I've done that by using developer tools in the browser. Then in the network tab setting the speed to "4G slow". Before this patch that results in grey nodes. After this patch it gives the correct color for the nodes.

Chris927 (approved):
It's an annoyingly ugly solution, as so repetitive. I cannot think of a better way, though, as the place where we could define anything shared is what we are trying to load here.

Awesome that the gray-node issue is resolved!

PR #499, opened 2026-06-02
fix(generator): remove duplicate Generator Start/Stop dropdown entry (#490)

### LINKED ISSUE #490: Secondary Generator start/stop with virtual generator
**Describe the bug**

Image

There is a secondary "_Generator Start/Stop_" in the dropdown of the generator input node after deploying a single virtual generator.

**Flow**
```
[{"id":"cf4bd9c2d21ecf59","type":"victron-virtual","z":"70bb18f661b2aeaa","name":"gen3","outputs":1,"device":"generator","default_values":true,"acload_nrofphases":1,"enable_s2support":false,"battery_capacity":25,"include_battery_temperature":false,"battery_voltage_preset":"24","battery_voltage_custom":24,"generator_type":"ac","generator_nrofphases":1,"include_engine_hours":false,"include_starter_voltage":false,"include_history_energy":false,"grid_nrofphases":1,"energymeter_nrofphases":1,"energymeter_role":"gridmeter","include_motor_temp":false,"include_controller_temp":false,"include_coolant_temp":false,"include_motor_rpm":true,"include_motor_direction":true,"position":0,"pvinverter_nrofphases":1,"switch_1_type":1,"switch_1_min":0,"switch_1_max":"","switch_1_initial":0,"switch_1_label":"","switch_1_unit":"","switch_1_step":1,"switch_1_customname":"","switch_1_group":"","switch_1_include_measurement":false,"switch_1_rgb_color_wheel":false,"switch_1_cct_wheel":false,"switch_1_rgb_white_dimmer":false,"switch_1_passthrough_mode":"auto_only","fluid_type":0,"include_tank_battery":false,"include_tank_temperature":false,"tank_battery_voltage":3.3,"tank_capacity":0.2,"auto_aggregate":false,"pulsemeter_multiplier":0.001,"ev_vin":"","ev_battery_capacity":"","temperature_type":2,"include_humidity":false,"include_pressure":false,"include_temp_battery":false,"temp_battery_voltage":3.3,"start_disconnected":false,"x":390,"y":120,"wires":[[]]},{"id":"cb94826e129f1ee2","type":"victron-input-generator","z":"70bb18f661b2aeaa","service":"","path":"","name":"","onlyChanges":false,"roundValues":"no","rateLimit":0,"conditionalMode":false,"condition1Operator":">","condition1HysteresisEnabled":false,"condition2Enabled":false,"condition2Service":"","condition2Path":"","condition2Operator":">","logicOperator":"AND","outputTrue":"true","outputFalse":"false","outputOnChange":false,"debounce":2000,"x":300,"y":280,"wires":[[]]},{"id":"d1e16509e82f0b38","type":"global-config","env":[],"modules":{"@victronenergy/node-red-contrib-victron":"1.7.13"}}]
```

dirkjanfaber:
When Venus OS creates a generator controller for a virtual genset, `ItemsChanged` signals can arrive during `_initService`'s async window before `GetValue` returns the device instance. With `deviceInstance=undefined`, `saveToCache` stores the entry under a no-trail key (e.g. `com.victronenergy.generator)` alongside the correct trail key (`com.victronenergy.generator/1`), and `getNodeServices` finds both, producing duplicate entries in the input-generator dropdown.

Fix: delete the stale no-trail cache entry whenever a trail entry is saved for the same service.

Also fix device instance `0` being treated as falsy in `getNodeServices`, which caused two controllers to both display as "Generator Start/Stop ()" instead of "Generator Start/Stop (0)" and "Generator Start/Stop (1)".

Add a Generator Start/Stop section to the virtual generator node docs explaining that Venus OS automatically creates a `com.victronenergy.generator` service and that input-generator/output-generator nodes should be used to interact with it.

Fixes https://github.com/victronenergy/node-red-contrib-victron/issues/490

### DIFF
--- a/src/services/victron-client.js
+++ b/src/services/victron-client.js
@@ -51,8 +51,8 @@ class VictronClient {
     const messageHandler = messages => {
       messages.forEach(msg => {
         _this.saveToCache(msg)
-        const trail = msg.deviceInstance == null ? '' : `/${msg.deviceInstance}`
-        const msgKey = `${msg.senderName}${trail}:${msg.path}`
+        const deviceInstanceSuffix = msg.deviceInstance == null ? '' : `/${msg.deviceInstance}`
+        const msgKey = `${msg.senderName}${deviceInstanceSuffix}:${msg.path}`
         debug(`[MESSAGE HANDLER] ${msgKey} | ${JSON.stringify(msg, null, 2)}`)
         // we remove msg.text, as we don't use it, and removing it makes it "in line" with what
         // we do in case of using regular callbacks from the cache, instead of polling (see
@@ -167,16 +167,16 @@ class VictronClient {
   saveToCache (msg) {
     let dbusPaths = {}
 
-    const trail = msg.deviceInstance == null ? '' : `/${msg.deviceInstance}`
+    const deviceInstanceSuffix = msg.deviceInstance == null ? '' : `/${msg.deviceInstance}`
 
-    if (this.system.cache[msg.senderName + trail]) { dbusPaths = this.system.cache[msg.senderName + trail] }
+    if (this.system.cache[msg.senderName + deviceInstanceSuffix]) { dbusPaths = this.system.cache[msg.senderName + deviceInstanceSuffix] }
 
-    // When a device instance is known, remove any stale no-trail entry for the same
+    // When a device instance is known, remove any stale no-suffix entry for the same
     // service. Such entries are created during _initService's async window (before
     // GetValue returns the device instance) when ItemsChanged signals arrive with
     // deviceInstance=undefined. Without this cleanup, both the stale entry and the
-    // correct trail entry match getNodeServices(), producing duplicate dropdown entries.
-    if (trail !== '') { delete this.system.cache[msg.senderName] }
+    // correct suffixed entry match getNodeServices(), producing duplicate dropdown entries.
+    if (deviceInstanceSuffix !== '') { delete this.system.cache[msg.senderName] }
 
     // some dbus messages are empty arrays [], and we interpret empty arrays as null values,
     // compare https://github.com/Chris927/dbus-native/commit/0080b9226a0ed9474be1e5ceeae58a9c78dfa046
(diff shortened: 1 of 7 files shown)


src/services/victron-client.js
Chris927: The value `msg.deviceInstance` will never end with a `/`, so this could be simplified to:

```js
const trail = msg.deviceInstance === null ? '' : `/${msg.deviceInstance`;
```

src/services/victron-client.js
```diff
@@ -167,6 +171,13 @@ class VictronClient {
 
     if (this.system.cache[msg.senderName + trail]) { dbusPaths = this.system.cache[msg.senderName + trail] }
 
+    // When a device instance is known, remove any stale no-trail entry for the same
+    // service. Such entries are created during _initService's async window (before
+    // GetValue returns the device instance) when ItemsChanged signals arrive with
+    // deviceInstance=undefined. Without this cleanup, both the stale entry and the
+    // correct trail entry match getNodeServices(), producing duplicate dropdown entries.
+    if (trail !== '') { delete this.system.cache[msg.senderName] }
+
```
Chris927, line 180: Suggestion: rename `trail` to `deviceInstanceSuffix` to clarify the intention of the value.

src/services/victron-client.js
```diff
@@ -167,6 +171,13 @@ class VictronClient {
 
     if (this.system.cache[msg.senderName + trail]) { dbusPaths = this.system.cache[msg.senderName + trail] }
 
+    // When a device instance is known, remove any stale no-trail entry for the same
+    // service. Such entries are created during _initService's async window (before
+    // GetValue returns the device instance) when ItemsChanged signals arrive with
+    // deviceInstance=undefined. Without this cleanup, both the stale entry and the
```
Chris927, line 177: Is it necessary to add to the cache, if `deviceInstance` is not given? If it is not necessary, then we could simplify this by not even adding to the cache in this case.

Chris927 (commented):
My concern here: These changes fix the issue at hand, and they aren't making things worse in the actual code. But: We can do some simplifications, and possibly get rid of the "trail" terminology, which hides what it really is: "trail" is the suffix used for cache entries, which is either empty, or it is the device Id.

So, it would be better to do some renaming. Otherwise, as an example, if anyone looks at the test labeled "stale no-trail entry alongside trail entry returns two services ...", they have to understand what "trail" and "no-trail" means; it is not clear from the context, and it is hard to explain: The name "trail" seems to come from the name of a local variable, which is based on the device Id of a victron service on the dbus. "trail" is an unnecessary, obfuscating abstraction, IMHO.

Chris927 (approved):

PR #501, opened 2026-06-03
fix(virtual-switch): default STEPPED DimmingMax to 7 when not configured

### LINKED ISSUE #500: Stepped switch when dropped to NR canvas starts with 10 segments, but 1 to 7 are allowed
**Describe the bug**
Stepped switch when dropped to NR canvas starts with 10 segments, but 1 to 7 are allowed.

dirkjanfaber:
Empty string default for switch_1_max caused `Number('') = 0`, setting DimmingMax to `0` on D-Bus. Replace `??` with `||` so falsy values (empty string, 0) fall back to `STEPPED_DEFAULT_MAX`.

The reported 10 segments could not be traced in the code; the root cause found was DimmingMax being set to `0`.

Fixes  https://github.com/victronenergy/node-red-contrib-victron/issues/500

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -957,6 +957,8 @@
     const savedType = context.switch_1_type !== undefined ? context.switch_1_type : victronVirtualConstantsExports.SWITCH_TYPE_MAP.TOGGLE;
     $('#node-input-switch_1_type').val(String(savedType));
 
+    let isInitialRender = true;
+
     function renderTypeConfig () {
       $('#switch-1-config-row').remove();
       $('#switch-1-pairs-row').remove();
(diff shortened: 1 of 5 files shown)


Chris927:
When testing manually: The config form loses its "max steps" setting, it seems. The below is what I see when I re-open the config form, after a deploy:

image

Further, the stepped virtual switch is not responding to me clicking any of the steps in the remote console. Not sure how to analyze this further.

resources/victron-virtual-functions.js
```diff
@@ -1510,10 +1514,16 @@
...
 	      $('#node-input-switch_1_type').closest('.form-row').after(configRow);
 
-	      // Restore saved values; on type change, skip type-specific fields
+	      // Restore saved values; on type change, skip type-specific fields.
+	      // Also restore if the rendered type matches the saved type (handles spurious change events
+	      // triggered by Node-RED's post-oneditprepare field population).
 	      cfg.fields.forEach(field => {
 	        const isCommonField = COMMON_SWITCH_FIELDS.some(f => f.id === field.id);
-	        const val = (isInitialRender || isCommonField) ? context[`switch_1_${field.id}`] : undefined;
+	        const typeUnchanged = Number(type) === Number(context.switch_1_type);
```
Chris927: [not critical] naming a boolean variable `is...` or `has...` helps to communicate through the name that it's a boolean. In this case, `isTypeUnchanged` would make it immediately clear that the variable contains a boolsn.

Chris927 (approved):
Looks good, confirmed it's working as intended, through manual testing.

PR #503, opened 2026-06-07
fix(gps): correct UtcTimestamp field name to UtcTime to align with GPS D-Bus specification

nmbath:
Summary
This PR fixes GPS UTC time path handling by replacing UtcTimestamp with UtcTime, matching the actual GPS D-Bus service definition and behavior.

Problem
The project expected the path UtcTimestamp for GPS UTC values, but the GPS service publishes UtcTime.
As a result, UTC time was not being received correctly from GPS services.

Changes
Updated GPS service whitelist from /UtcTimestamp to /UtcTime
Updated missing path metadata to use /UtcTime and corrected description
Added /UtcTime to GPS service definitions so it is available in node selection
Updated virtual GPS example output to include UtcTime as milliseconds since midnight UTC
Why this is correct
In the dbus_gps implementation, UtcTime is the active published field.
UtcTimestamp exists in source only as disabled code, and is not published.

Testing
Verified branch contains the expected changes for GPS UTC path handling
Confirmed virtual GPS example now emits UtcTime in payload
Impact
Backward-compatible for users not using UTC path
Fixes UTC path consumption for real GPS and virtual GPS flows


dirkjanfaber:
Closing in favour of #505

PR #504, opened 2026-06-08
feat(virtual-acload): add S2 support with configurable power measurement

dirkjanfaber:
- Add Connect/Disconnect/Message/KeepAlive D-Bus handlers for S2 sessions
- Restore PowerMeasurementStart/Stop handling and emitCallback snapshot emission
- Add s2_measurement_type config: 3-phase symmetric, per-phase (L1/L2/L3), or single-phase on L1/L2/L3
- Show 'S2 communication' label on port 2 when S2 is enabled
- Show/hide measurement type dropdown based on S2 checkbox state
- Reset power measurement state on Connect/Disconnect

Note that this isn't easy to test. Please just check the code if there isn't anything obviously wrong.

### DIFF
--- a/src/nodes/victron-virtual.html
+++ b/src/nodes/victron-virtual.html
@@ -89,57 +89,7 @@
             return (this.name||"Virtual "+this.device)
         },
         outputLabels: function(index) {
-            // Inline logic since window.getOutputLabels may not be loaded yet
-            if (this.device === 'pulsemeter') {
-                return index === 0 ? 'Passthrough' : 'Aggregate'
-            }
-            if (this.device === 'acload' && this.enable_s2support) {
-                return index === 0 ? 'Passthrough' : 'S2 communication'
-            }
-            if (!this.device || this.device !== 'switch') {
-                return index === 0 ? 'Passthrough' : 'Output ' + (index + 1)
-            }
-
-            const switchType = this.switch_1_type
-            const typeKey = switchType !== undefined ? parseInt(switchType, 10) : 1 // TOGGLE = 1
-
-            // Map of second output labels (for switches that don't have "State")
-            const secondLabels = {
-                3: 'Temperature',  // TEMPERATURE_SETPOINT
-                6: 'Selected',     // DROPDOWN
-                7: 'Value'         // BASIC_SLIDER
-            }
-
-            // Map of third output labels (only for switches with 3 outputs)
-            const thirdLabels = {
-                2: 'Dimming',      // DIMMABLE
-                4: 'Value',        // STEPPED
-                8: 'Value',        // NUMERIC_INPUT
-                11: 'LightControls', // RGB_COLOR_WHEEL
-                12: 'LightControls', // CCT_WHEEL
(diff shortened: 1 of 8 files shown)


src/nodes/victron-virtual/device-type/acload.js
```diff
@@ -68,11 +68,23 @@ function initialize (config, ifaceDesc, iface, node) {
...
+      L1_L2_L3: {
+        'Ac/L1/Power': 'ELECTRIC.POWER.L1',
+        'Ac/L2/Power': 'ELECTRIC.POWER.L2',
+        'Ac/L3/Power': 'ELECTRIC.POWER.L3'
+      },
+      L1: { 'Ac/L1/Power': 'ELECTRIC.POWER.L1' },
+      L2: { 'Ac/L2/Power': 'ELECTRIC.POWER.L2' },
+      L3: { 'Ac/L3/Power': 'ELECTRIC.POWER.L3' }
     }
+    ifaceDesc.__s2PowerMeasurementProps = measurementPropsMap[measurementType] || measurementPropsMap['3_PHASE_SYMMETRIC']
```
Chris927: The logical or is unnecessary, you can remove `|| measurementPropsMap['3_PHASE_SYMMETRIC']`. We ensure in line 71 that `measurementType` always has a value, and if the value is invalid, then we have a coding error somewhere, and this logical or would hide that.

src/nodes/victron-virtual/index.js
```diff
@@ -194,21 +194,26 @@ module.exports = function (RED) {
...
-      }
-      const outputs = [msg]
-      // Fill remaining outputs with null
-      for (let i = 1; i < config.outputs; i++) {
-        outputs.push(null)
-      }
-      node.send(outputs)
+      // S2 signal messages are internal control - never pass through to any output
+      if (msg.payload && msg.payload.s2Signal !== undefined) {
+        // Fall through to s2Signal handler below
```
Chris927, line 199: This is a semantic change, which is not backwards-compatible. This is likely the intention, I am calling it out, though, to ensure it's indeed intentional :)

src/nodes/victron-virtual.html
```diff
@@ -92,6 +93,9 @@
             if (this.device === 'pulsemeter') {
                 return index === 0 ? 'Passthrough' : 'Aggregate'
             }
+            if (this.device === 'acload' && this.enable_s2support) {
```
Chris927: [not critical, refactoring idea]: This is brittle logic, which we cannot unit-test, as it lives in the html file. however, we can use functions from `victron-virtual-functions.js` here, right? I suggest we move this logic into a function `determineOutputLabels(node, index)` or similar, in `victron-virtual-functions.js`, and then cover all paths with unit tests.

Chris927 (approved):
Added some non-critical comments. As suggested, I didn't even try to test manually :)

Maybe create an issue to remind us of the refactoring we can do here?

src/nodes/victron-virtual/index.js
dirkjanfaber: Acknowledged.

PR #506, opened 2026-06-15
feat(virtual-switch): add live SVG preview for all switch types

### LINKED ISSUE #502: [Feature] Add SVG preview to virtual switches
Just as #496 for the virtual indicators, but then for the virtual switches.

dirkjanfaber:
Adds a live SVG card preview in the node edit dialog for all virtual switch types (toggle, momentary, dimmable, temperature setpoint, stepped, dropdown, basic slider, numeric input, three-state, bilge pump, RGB controls). The preview updates in real time as name, group, and type-specific fields are changed.

Preview layout matches the GX device panel style: group icon, group/type label, switch name, and a type-specific control strip. Each control type renders a representative interactive state (e.g. Off|On toggle, stepped segments with Off, brightness slider with position indicator, color swatch for RGB). Colors and layout are extracted into named constants.

Closes #502


Chris927 (approved):
Great visual enhancement! Not blocking, but maybe worth considering:

- `resources/victron-virtual-functions.js` increased from 93k to 115k. Is disk space a concern on the Cerbo still? If it is, we may want to be careful with making changes that result in a lot of code (thanks to AI coding, I guess).
- In the config dialog I noticed that, when switching to a different switch type, the values for `name` and `group` get reset. Is this intentional? If not, we may want to treat it as an issue.

dirkjanfaber:
> * `resources/victron-virtual-functions.js` increased from 93k to 115k. Is disk space a concern on the Cerbo still? If it is, we may want to be careful with making changes that result in a lot of code (thanks to AI coding, I guess).

This made me realise that I hadn't removed the old `.svg` files yet, which account for a total of more than the functions file difference. 
And space is still a concern (not a huge one right now), but saving things where possible is still wise.

I will update the PR to remove the no longer needed files.

> * In the config dialog I noticed that, when switching to a different switch type, the values for `name` and `group` get reset. Is this intentional? If not, we may want to treat it as an issue.

I don't recall that on my system. Hm... I'll double-check.

dirkjanfaber:
>> In the config dialog I noticed that, when switching to a different switch type, the values for name and group get reset. Is this intentional? If not, we may want to treat it as an issue.
> I don't recall that on my system. Hm... I'll double-check

Unable to reproduce on my system. So if there actually is a problem, lets fix that via a new issue.

PR #509, opened 2026-06-16
fix(nodes): prevent crash and guide recovery when config node is missing

### LINKED ISSUE #492: AI generated Victron nodes, crash Node-RED
**Describe the bug**
AI generated Victron nodes seems to crash NR.

**To Reproduce**

Import the AI generated flow: [gemini-code-1779974795578.json](https://github.com/user-attachments/files/28349870/gemini-code-1779974795578.json)
Make the changes per this doc: [Doc1.docx](https://github.com/user-attachments/files/28349869/Doc1.docx)

**Expected behavior**
NR should crash.

**Flow**
Then, set NR to safe mode, delete the Victron nodes, drag and set them from scratch.  Deploy and set NR back to enabled, and all works.

Cerbo GX.

 - Venus OS version: 3.73

dirkjanfaber:
When a flow is imported without a victron-client-config node (e.g. an AI-generated flow), the `onadd` hook that auto-creates one does not fire. Both `BaseInputNode` and `BaseOutputNode` previously dereferenced `null.client` immediately, crashing Node-RED.

Now they log an error and show a red '_no config_' status instead, leaving the node inert without taking down the runtime. The error message tells the user exactly how to recover: drag any Victron node from the palette (triggers `onadd`, which auto-creates the config node), then deploy.

Closes #492

### DIFF
--- a/src/nodes/victron-nodes.js
+++ b/src/nodes/victron-nodes.js
@@ -43,6 +43,15 @@ module.exports = function (RED) {
     }
   }
 
+  const isConfigNodePresent = (node) => {
+    if (node.configNode) {
+      return true
+    }
+    node.error('No Victron client config node found. Drag any Victron node from the palette to auto-create it, then deploy.')
+    node.status({ fill: 'red', shape: 'ring', text: 'no config' })
+    return false
+  }
+
   // Helper function to format status text consistently
   const formatStatusText = (value, suffix) => {
     const valueText = value !== null && value !== undefined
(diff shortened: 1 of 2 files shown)


src/nodes/victron-nodes.js
```diff
@@ -467,6 +472,11 @@ module.exports = function (RED) {
       this.initialValue = initialValue
 
       this.configNode = RED.nodes.getNode('victron-client-id')
+      if (!this.configNode) {
+        this.error('No Victron client config node found. Drag any Victron node from the palette to auto-create it, then deploy.')
```
Chris927: The change is basically copy&paste. Rather refactor to:

At top of file:

```js
function isConfigNodePresent(node) {
  if (node.configNode) {
    return true;
  } else {
    node.error(...)
    node.status(...)
  }
}
```

In the node constructors:

```js
if (!isConfigNodePresent(this)) {
  return // we cannot continue
}
```

Chris927 (changes requested):
Rather make it DRY.

src/nodes/victron-nodes.js
Chris927: Update: there is a `return false;` missing in the `else` of my `isConfigNodePresent()`...

Chris927 (approved):

PR #511, opened 2026-06-16
fix(dbus-listener): singleton service cache keys, null suppression, pathObj guards

### LINKED ISSUE #346: Solar charger node sends null
**Describe the bug**
Via https://community.victronenergy.com/t/venus-os-v3-70-61-available-for-public-testing/45708/239

> There is a bug in Solar Charger node in Node Red.
> 
> The node is configured to read Relay state from Smart Solar 250/80 TR VE.Can - only changes.
> Relay on mppt is configured to “Day detection (panels irradiated)” mode.
> The node should send `0` or `1` in related conditions, but sometimes it sends null.
> It happens at least once a day with any connection to real panels irradiation.
>
> Now I am on Venus 3.70 ~60. The same issue was also noticed in all 3.6X as I remember.
> Before 3.6 the same flow was working like a charm.

> I attached a picture with all required information and debug log.

**To Reproduce**
1. Using node  '...'
3. See error

**Expected behavior**
No `null` output

**Screenshots**

Image

**Flow**
If applicable, add a flow to help explain your problem.

 - Hardware: [e.g. Cerbo GX, Ekrano, Raspberry Pi]

 - node-red-contrib-victron version:
 - Venus OS version:
 

**Additional context**

dirkjanfaber:
## Summary

- **Singleton service cache keys** (`0810cd9`, `8eb62d2`, `bd67dc8`): Services without a device instance (system, settings, platform, dynamicess) were previously keyed inconsistently in the cache, causing duplicate dropdown entries and stale lookups. Now uses a plain key (no `/0` suffix) for singleton services, with migration for previously stored service names.

- **Suppress null values from reaching node subscribers** (`85628d8`): During service re-initialization (e.g. a solar charger reconnecting at sunrise), `_requestRoot` sent GetItems results including paths whose D-Bus value is temporarily null. These bypassed the `onlyChanges` filter and were emitted as spurious null payloads to flows. Two complementary guards added - one in `_requestRoot` for explicit null, one in `messageHandler` after `saveToCache` converts Victron's empty-array null encoding - making all three signal paths (`PropertiesChanged`, `ItemsChanged`, `_requestRoot`) consistent. Null values are still saved to cache so paths appear in the dropdown with their `- (null)` label. Fixes #346.

- **Guard against undefined pathObj** (`4474922`): Input and output nodes crashed on the first received message when `pathObj` was undefined (legacy flows, unconfigured nodes, imported flows without a path selected). Input nodes now skip enum formatting and still emit the raw value; output nodes show a red status and skip the write.

### DIFF
--- a/src/nodes/victron-nodes.js
+++ b/src/nodes/victron-nodes.js
@@ -1,20 +1,20 @@
 const { debounce } = require('../services/utils.js')
+const { serviceNamesWithoutDeviceInstance } = require('../services/dbus-listener.js')
 
-/*
- * If a service name in the Node-RED node differs from the actual service name used
- * in the dbus, we can define a mapping here.
- * We need this for now for system input nodes: Previously, they had '/0' appended,
- * but due to PR#511 (https://github.com/victronenergy/node-red-contrib-victron/pull/511)
- * their service name is now just 'com.victronenergy.system'.
+/* Those service names which should not have a device instance any more may still have
+ * a device instance in a pre-existing node definition. We therefore strip the device
+ * instance from the service name for those services.
+ * See PR#511 (https://github.com/victronenergy/node-red-contrib-victron/pull/511)
+ * for details.
  */
-const serviceNameMappings = {
-  'com.victronenergy.system/0': 'com.victronenergy.system'
+function mustRemoveDeviceInstanceFromService (service) {
+  const serviceNameBare = service.split('/')[0] // Remove any device instance suffix
+  return serviceNamesWithoutDeviceInstance.includes(serviceNameBare)
 }
 
 module.exports = function (RED) {
   const debug = require('debug')('node-red-contrib-victron:victron-nodes')
   const utils = require('../services/utils.js')
-  const { serviceNamesWithoutDeviceInstance } = require('../services/dbus-listener.js')
 
   const migrateSubscriptions = (x) => {
     // Check if client is fully initialized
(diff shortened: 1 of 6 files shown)


dirkjanfaber:
Update: This could break some flows as we now change services from eg. `com.victronenergy.system/0` to `com.victronenergy.system`.
That needs a proper fix too.

src/nodes/victron-nodes.js
```diff
@@ -1,5 +1,16 @@
 const { debounce } = require('../services/utils.js')
 
+/*
+ * If a service name in the Node-RED node differs from the actual service name used
+ * in the dbus, we can define a mapping here.
+ * We need this for now for system input nodes: Previously, they had '/0' appended,
+ * but due to PR#511 (https://github.com/victronenergy/node-red-contrib-victron/pull/511)
+ * their service name is now just 'com.victronenergy.system'.
+ */
+const serviceNameMappings = {
```
dirkjanfaber: I think we should reuse `serviceNamesWithoutDeviceInstance` (from https://github.com/victronenergy/node-red-contrib-victron/blob/8eb62d2f4b0bace4f3b4c0b998d4aaf861ad3a61/src/services/dbus-listener.js#L90).

There are currently 4 entries there and I am quite sure we need all of those.

dirkjanfaber (commented):
See the comment.

dirkjanfaber:
Closing this as I've split the changes up into 3 seperate PR's: #517 , #518 & #519. 

To make it better traceable in case we unintentionally happen to add a regression.

PR #514, opened 2026-06-22
feat(pvinverter): auto-calculate Ac/Energy/Forward from power

dirkjanfaber:
When enabled, the virtual PV inverter integrates power over time to accumulate `Ac/Energy/Forward`. For multi-phase setups the total is the sum of per-phase energies; falls back to `Ac/Power` when no per-phase data is provided. Value persists across restarts.

Configurable via checkbox; defaults to on for new nodes. Existing flows treat the absent field as disabled (safe migration).

Also adds `flushPersistedState()` to `persist.js`: cancels any pending throttle timer and writes immediately. Called on node close so throttled-persist state (e.g. accumulated energy) is not lost on redeploy.

### DIFF
--- a/src/nodes/victron-virtual.html
+++ b/src/nodes/victron-virtual.html
@@ -479,7 +479,9 @@
             </div>
             <div class="form-row victron-checkbox">
                 <input type="checkbox" id="node-input-pvinverter_auto_energy">
-                <label for="node-input-pvinverter_auto_energy">Auto-calculate energy (Ac/Energy/Forward)</label>
+                <label for="node-input-pvinverter_auto_energy">Auto-calculate energy (Ac/Energy/Forward)
+                    <i class="fa fa-info-circle tooltip-icon" data-tooltip="Integrates power over time to estimate energy. Precision improves with higher update frequency."></i>
+                </label>
             </div>
         </div>
 
--- a/src/nodes/victron-virtual.html
+++ b/src/nodes/victron-virtual.html
@@ -38,6 +38,7 @@
             // pvinverter
             position: { value: 0, required: false},
             pvinverter_nrofphases: {value: 1, required: false},
+            pvinverter_auto_energy: {value: true, required: false},
             // switch
             switch_1_type: {value: 1, required: false},
             switch_1_min: {value: 0, required: false},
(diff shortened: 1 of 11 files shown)


src/nodes/victron-virtual/device-type/pvinverter.js
```diff
@@ -70,4 +80,46 @@ function initialize (config, ifaceDesc, iface, node) {
   return `Virtual ${iface.NrOfPhases}-phase pvinverter`
 }
 
-module.exports = { properties, initialize, label: 'PV inverter' }
+function accumulateDelta (changes, instance, energyKey, oldPower, lastTs, now) {
+  if (lastTs != null && oldPower != null && oldPower > 0 && !(energyKey in changes)) {
```
Chris927: The condition `oldPower > 0` is suspect: values for power can be positive and negative, but when accumulating, we only accumulate positives. (Admittedly, negative power may not make sense for a PV inverter, but allowing negative power but then not accumulating it is inconsistent.)

Suggestion: either accumulate both positive and negative power, or add validation to the various power inputs to disallow negative power.

src/nodes/victron-virtual.html
```diff
@@ -476,6 +477,10 @@
                     <option value="3">3</option>
                 </select>
             </div>
+            <div class="form-row victron-checkbox">
+                <input type="checkbox" id="node-input-pvinverter_auto_energy">
+                <label for="node-input-pvinverter_auto_energy">Auto-calculate energy (Ac/Energy/Forward)</label>
```
Chris927: Energy calculated will not be precise: If there are energy spikes in-between measurements, or if the frequency of updates is too low to capture fluctuations, then the energy calculated will be far off. Example: I measure every 5 minutes, at the time of measuring the power there is a cloud covering the PV panels, but in between it is fully sunny.

Is it useful or necessary to mention this / warn the user about it?

Chris927 (changes requested):
The issue around positive / negative power should be addressed, the rest is not blocking IMHO.

A question, though: Why implement this for PV inverter only, why not also for energy meters etc?

src/nodes/victron-virtual.html
```diff
@@ -476,6 +477,12 @@
                     <option value="3">3</option>
                 </select>
             </div>
+            <div class="form-row victron-checkbox">
+                <input type="checkbox" id="node-input-pvinverter_auto_energy">
+                <label for="node-input-pvinverter_auto_energy">Auto-calculate energy (Ac/Energy/Forward)
+                    <i class="fa fa-info-circle tooltip-icon" data-tooltip="Integrates power over time to estimate energy. Precision improves with higher update frequency."></i>
```
Chris927, line 483: I like it, it's a clever way of saying "it's not that precise" ;)

Chris927 (approved):

Issue #513, opened 2026-06-21
[Feature] Virtual AC Load / Virtual Device: persistent state for energy counters after Node-RED restart


alexkn:
**Is your feature request related to a problem? Please describe.**
When using the node-red-contrib-victron Virtual AC Load (or other Virtual Devices), any runtime-calculated values such as accumulated energy counters are lost after a Node-RED restart.
In my setup, I use a Virtual AC Load to represent a heating element. The device only exposes power (W), so I calculate energy (kWh) in Node-RED by integrating power over time and send it to /Ac/Energy/Forward
However, after a Node-RED restart, this value is reset because the integration state is not restored.

**Describe the solution you'd like**
I would expect that Virtual Devices persist the energy values over restarts.

alexkn:
@dirkjanfaber @Chris927 
Would it be possible to merge it into version 1.6 and release it for Venus 3.7x?

dirkjanfaber:
No, the code has diverted that much that the backport isn't trivial. And 3.8x is due in a couple of weeks from now. Switching to the beta branch is the best option to get this earlier if you need it already.

PR #515, opened 2026-06-22
fix(virtual): persist energy counters across restarts for all device types

dirkjanfaber:
Virtual AC load, grid meter, energy meter, and genset energy counters (`Ac/Energy/Forward`, `Ac/Energy/Reverse`, `Ac/L*/Energy/Forward`, `Ac/L*/Energy/Reverse`, `History/EnergyOut`) are now persisted with a 60s throttle. Values survive Node-RED restarts regardless of whether they are set externally or auto-calculated.

Also extends the pvinverter energy persist introduced in #514 to be unconditional: previously persist was only enabled when `pvinverter_auto_energy` was `true`, leaving manually-set energy values unprotected on restart. All device types now persist energy counters on the same basis.

Fixes https://github.com/victronenergy/node-red-contrib-victron/issues/513

### DIFF
--- a/src/nodes/victron-virtual/device-type/acload.js
+++ b/src/nodes/victron-virtual/device-type/acload.js
@@ -18,8 +18,8 @@ const phaseProperties = [
   { name: 'Current', unit: 'A' },
   { name: 'Power', unit: 'W' },
   { name: 'Voltage', unit: 'V' },
-  { name: 'Energy/Forward', unit: 'kWh' },
-  { name: 'Energy/Reverse', unit: 'kWh' },
+  { name: 'Energy/Forward', unit: 'kWh', persist: ENERGY_PERSIST_SECONDS },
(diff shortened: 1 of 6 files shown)


src/nodes/victron-virtual/device-type/acload.js
```diff
@@ -44,10 +46,14 @@ function initialize (config, ifaceDesc, iface, node) {
     const phase = `L${i}`
     phaseProperties.forEach(({ name, unit }) => {
       const key = `Ac/${phase}/${name}`
-      ifaceDesc.properties[key] = {
+      const propDef = {
         type: 'd',
         format: (v) => v != null ? v.toFixed(2) + unit : ''
       }
+      if (name.startsWith('Energy/')) {
```
Chris927: [not blocking] instead of having logic, each element in the `phaseProperties` array could have an option `persist: ENERGY_PERSIST_SECONDS` field.

Chris927 (approved):
Nothing blocking, just a comment.

PR #516, opened 2026-06-23
fix(virtual-switch): prevent passthrough output from receiving mutated payload

dirkjanfaber:
`msg.payload` was reassigned to the expanded object after `node.send()` was called, but Node-RED delivers the message asynchronously, so downstream nodes received the expanded form instead of the original payload.

Remove the mutation - use the expanded variable directly throughout handleInput instead of assigning it back to `msg.payload`.

### DIFF
--- a/src/nodes/victron-virtual-switch.js
+++ b/src/nodes/victron-virtual-switch.js
@@ -62,9 +62,8 @@ module.exports = function (RED) {
         done()
         return
       }
-      msg.payload = expanded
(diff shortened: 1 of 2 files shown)


Chris927 (approved):
I've tested the code path manually, all good.

Note however: The unit test is possibly useful, but not related to the code change: The changed code currently has no code coverage, which we should address, compare suggested discussion points.

PR #518, opened 2026-06-24
fix(dbus-listener): use plain cache key for singleton services

dirkjanfaber:
Services without a device instance (settings, platform, system, dynamicess) were keyed as 'service/0' in the cache, causing duplicate dropdown entries and stale lookups. Switch to the plain service name as the cache key.

Migrates previously stored '/0'-suffixed entries in flows deployed with the old key format.

### DIFF
--- a/src/services/dbus-listener.js
+++ b/src/services/dbus-listener.js
@@ -274,15 +274,15 @@ class VictronDbusListener {
           service.fluidType = data.FluidType
         }
 
-        const isWithoutDeviceInstance = serviceNamesWithoutDeviceInstance.includes(service.name)
+        const needsDeviceInstance = !serviceNamesWithoutDeviceInstance.includes(service.name)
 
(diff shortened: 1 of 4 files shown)


src/services/dbus-listener.js
```diff
@@ -269,11 +274,15 @@ class VictronDbusListener {
           service.fluidType = data.FluidType
         }
 
-        if (!service.deviceInstance && data['/DeviceInstance'] != null) {
+        const isWithoutDeviceInstance = serviceNamesWithoutDeviceInstance.includes(service.name)
```
Chris927: [not blocking] If we turn the boolean around, the code below would avoid double negation:

```js
const needsDeviceInstance = !serviceNamesWithoutDeviceInstance.includes(service.name)
```

Chris927 (approved):
Other than a suggested variable name change this looks good, and my manual tests were green.

PR #520, opened 2026-06-24
feat(ev): make Mgmt/Connection writable with auto-set from EVCS device instance

dirkjanfaber:
- Add Mgmt/Connection as writable ~~integer~~ string (second commit) path for virtual EV device
- Auto-set value from EVCS device instance when AtSite is selected
- Guard config block on ev_evcs_device_instance presence
- Extract hasEvcsInstance helper to reduce duplication

These changes are needed to allow the charging of the virtually attached EV to be controllable via Dynamic ESS.


src/nodes/victron-virtual-functions.js
```diff
@@ -971,6 +971,28 @@ function renderDropdownLabels (context) {
...
+  return fetch(url)
+    .then(response => response.json())
+    .then(data => {
+      const result = []
+      for (const service in data) {
+        if (service.indexOf('com.victronenergy.evcharger') === 0) {
+          const paths = data[service]
+          const deviceInstance = paths['/DeviceInstance'] != null
+            ? paths['/DeviceInstance']
+            : parseInt(service.split('/')[1], 10)
```
Chris927, line 985: second parameter `10` is the default, can be removed.

src/nodes/victron-virtual/device-type/ev.js
```diff
@@ -70,6 +73,10 @@ function onPropertiesChanged ({ changes, instance }) {
     changes['LastUpdated/Charging'] = now
   }
 
+  if ('AtSite' in changes && hasEvcsInstance(config)) {
```
Chris927: Should there also be `... && changes['Mgmt/Connection'] === undefined` in the condition? Otherwise, `Mgmt/Connection` would not be writable if `hasEvcsInstance(config)` is true, I believe. But maybe that's the intention?

Chris927 (changes requested):
I've got one concern about `Mgmt/Connection` being writable, see below. Looks good otherwise.

Chris927 (approved):

PR #524, opened 2026-06-26
Add end-to-end testing

Chris927:
Adds end-to-end testing using testcafe.

The tests can be run locally against a Venus OS device, provided the Node-RED http endpoint and mqtt are accessible. See the README for details.

The tests also run on every push to the master branch.

The scope of these end-to-end tests is being able to catch regressions which cannot be protected against through unit testing, e.g. changes related to the html files, the deployment of nodes, and nodes communicating with dbus.

Not every node type is being tested at this point, so we should expand on these tests.


e2etest/setup-node-red.js
```diff
@@ -0,0 +1,135 @@
...
+    })
+    req.on('error', reject)
+    if (body) req.write(body)
+    req.end()
+  })
+}
+
+function nodeRedRequest (proxyDomain, method, path, sessionCookie, body, extraHeaders) {
+  const bodyStr = body ? JSON.stringify(body) : undefined
+  console.log(`Making Node-RED request: ${method} ${path}, body=${bodyStr}, extraHeaders=${JSON.stringify(extraHeaders)}, proxyDomain=${proxyDomain}, sessionCookie=${sessionCookie}`)
```
dirkjanfaber: I think we should remove the actual cookie from this log line.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,80 @@
...
+          sleep 15
+
+      - name: List browsers available for E2E tests
+        run: npm run test:e2e -- --list-browsers
+
+      - name: Run E2E tests
+        env:
+          VRM_API_TOKEN: ${{ secrets.VRM_API_TOKEN }}
+          VRM_INSTALLATION_ID: ${{ secrets.VRM_INSTALLATION_ID }}
+          SKIP_MQTT_VERIFICATION: 'true'
```
dirkjanfaber: I don't believe we are using `SKIP_MQTT_VERIFICATION` in the code anymore.

e2etest/mqtt-verify.js
```diff
@@ -0,0 +1,226 @@
...
+//   N/{portalId}/settings/0/Settings/Devices/{suffix}/ClassAndVrmInstance
+// which publishes a string like "switch:0" (class:instance).
+//
+// Required env vars: VRM_API_TOKEN
+// Optional env vars: VRM_PORTAL_ID (default c0619ab4e242), VRM_MQTT_USERNAME (VRM email)
+
+const https = require('https')
+const mqtt = require('mqtt')
+
+const VRM_PORTAL_ID = process.env.VRM_PORTAL_ID || 'c0619ab4e242'
```
dirkjanfaber: I don't like the hardcoded ID.
Better to throw an error when not set. E.g. `if (!VRM_PORTAL_ID) throw new Error('VRM_PORTAL_ID env var is required')`

e2etest/vrm-auth.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
+
+const https = require('https')
+
+const VRM_INSTALLATION_ID = process.env.VRM_INSTALLATION_ID || '445836'
```
dirkjanfaber: Another hardcoded ID we should remove.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,80 @@
...
+        run: npm run test:e2e
+
+      # We do not teardown, so that we can see the state of Node-RED and the flows after the test run.
+      # - name: Teardown Node-RED
+      #   if: always()
+      #   env:
+      #     VRM_API_TOKEN: ${{ secrets.VRM_API_TOKEN }}
+      #     VRM_INSTALLATION_ID: ${{ secrets.VRM_INSTALLATION_ID }}
+      #   run: node e2etest/teardown-node-red.js
```
dirkjanfaber: Not so certain that we shouldn't teardown. For debugging yes, it is handy to have it. But it makes a secondary run different from the initial one.

src/nodes/victron-virtual-switch.js
```diff
@@ -128,16 +128,24 @@ module.exports = function (RED) {
     function instantiateDbus (self) {
       debug('instantiateDbus called for node:', self.id)
       // Connect to the dbus
+      function createClientCallback (err, bus) {
```
dirkjanfaber: I don't think we actually use `bus` here, do we?

README.md
```diff
@@ -64,6 +64,30 @@ This library is community-supported. For help:
 
 Contributions are welcome! See our [Contributing Guide](https://github.com/victronenergy/node-red-contrib-victron/blob/master/CONTRIBUTING.md) for details.
 
+## Development
+
+### Setup
+
+TBD
```
dirkjanfaber, line 71: Probably needs some extra text.

dirkjanfaber (changes requested):
Really nice to get to this stage of having E2E in. Still some minor findings.

e2etest/setup-node-red.js
Chris927: Done.

.github/workflows/e2e.yml
Chris927: I've moved the teardown of the flows to the setup, see 968cce3. This should (a) ensure that the runs always start without flows, and (b) that the flows are still available after the test run.

README.md
Chris927: I suggest to do this outside of this PR, I created [this issue](https://github.com/victronenergy/node-red-contrib-victron/issues/529) for it.

e2etest/setup-node-red.js
```diff
@@ -94,6 +86,11 @@ async function installPackage (proxyDomain, sessionCookie) {
 
 async function main () {
   const { sessionCookie, proxyDomain } = await fetchSessionCookie()
+  try {
+    await clearAllFlows(proxyDomain, sessionCookie)
+  } catch (err) {
+    console.warn('Error clearing flows, continuing anyway:', err.message)
```
dirkjanfaber: I think this is too easy to miss if it happens. Github Actions offers a way to set a warning message: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-a-warning-message
We can use that by doing something like:
```
process.stdout.write(`::warning file=e2etest/setup-node-red.js,line=92::Error clearing flows (continuing anyway): ${err.message}\n`)
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,82 @@
...
+        run: npm run test:e2e -- --list-browsers
+
+      - name: Run E2E tests
+        env:
+          VRM_API_TOKEN: ${{ secrets.VRM_API_TOKEN }}
+          VRM_INSTALLATION_ID: ${{ secrets.VRM_INSTALLATION_ID }}
+          VRM_PORTAL_ID: ${{ secrets.VRM_PORTAL_ID }}
+        run: npm run test:e2e
+
+      # We do not teardown, so that we can see the state of Node-RED and the flows after the test run.
```
dirkjanfaber: This comment can go now, right?

dirkjanfaber (changes requested):
I am tempted to approve it. But still requesting 2 more (minor) changes before doing that.

dirkjanfaber (approved):
Super to get this in! Sure we will need some changes in the coming weeks.

Issue #512, opened 2026-06-18
Long name entries aren't readable in dropdowns


dirkjanfaber:
**Describe the bug**

<img width="966" height="462" alt="Image" src="https://github.com/user-attachments/assets/08bd179d-de49-4816-b1ab-830dad83117d" />

It looks like the max length is about 35 characters. And we have quite a lot of entries in the `services.json` file that are longer. 

Longest is:
```
        "name": "Set it before a manual start to make the generator run for a fixed amount of seconds. Once the generator is started it indicates the seconds left until the generator is stopped",
```

PR #526, opened 2026-06-27
fix(nodes): replace path select with searchable combobox (#512)

dirkjanfaber:
Long measurement names were truncated in the native <select> dropdown, making them unreadable. Replace the Measurement <select> with a custom combobox: a text input filters the list as you type, and the dropdown shows full wrapping text for every entry. The hidden <select> is kept for Node-RED value storage and auto-save compatibility.

https://github.com/victronenergy/node-red-contrib-victron/issues/512

### DIFF
--- a/resources/victron-styles.css
+++ b/resources/victron-styles.css
@@ -769,6 +769,63 @@
   color: #dc3545;
 }
 
+/* Path combobox */
+.victron-form .victron-path-container {
+  display: inline-block;
+  width: 240px;
+  vertical-align: top;
+  position: relative;
+}
+
+.victron-form .victron-combobox-input {
+  width: 100% !important;
+  box-sizing: border-box;
+}
+
+.victron-form .victron-combobox-list {
+  display: none;
+  position: absolute;
+  top: 100%;
+  left: 0;
+  min-width: 100%;
+  max-height: 200px;
+  overflow-y: auto;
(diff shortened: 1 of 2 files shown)


Chris927 (changes requested):
This PR is implementing a custom combo box in untestable code: It works mostly, but there are scenarios where you cannot get the options back without clicking outside of the input field.

Suggestion: Use an alternative entirely based on CSS, to avoid the complexity and brittleness of untestable JS/jQuery code: Applying the below CSS to the `input` element should work:

```css
word-wrap: break-word;
white-space: normal;
min-height: 48px;
```

The downside is that the height is larger than necessary in cases the text does not have to wrap. I think this is acceptable. (Maybe this can be improved with clever CSS?)

The upside is that we keep using the native `select` element, not risking our logic breaking in browsers we do not actively test. And obviously we avoid a lot of (currently) untestable code.

Another upside: If there is a browser that doesn't support any of the css, it will degrade to the current behavior (the end of the text cut off), and still work.

If the downside (unnecessary height) is unacceptable, then I suggest using something like [Select2](https://select2.org/) from a CDN: Our JavaScript would be limited to a function call, and it would again degrade gracefully to the native `select` behavior.

dirkjanfaber:
Update; Lets go for select2 instead.

PR #527, opened 2026-06-28
fix(virtual): sanitize node ID for D-Bus paths to support subflows

### LINKED ISSUE #525: Victron Energy nodes in subflows are not published to the dbus
Unfortunately the Victron Energy nodes currently do not work in subflows. This makes it cumbersome to create things like repeating groups of switches with a certain logic without having to duplicate nodes and/or code.

**To Reproduce**
1. Monitor the Venus OS GUI or `dbus-spy`
2. Go to Node-RED and create a subflow
3. Add a Victron Energy `Virtual Switch` node and deploy
4. There is no new dbus entity added

**Hardware**
 - Hardware: Multiplus II GX

**Software**
 - node-red-contrib-victron version: 1.6.64
 - Venus OS version: 3.75

dirkjanfaber:
Node-RED generates subflow node IDs as `${instanceId}-${templateId}` introducing hyphens (and dots from old-format IDs) that are invalid in D-Bus object path elements. The Venus OS settings daemon rejects `AddSettings` calls for paths like `/Settings/Devices/virtual_abc-def/...`, causing 'Virtual switch setup failed' for any node placed in a subflow.

Replace any non-alphanumeric character with `_` before building the D-Bus service name and settings path. Persistence file names keep the original node ID since the filesystem accepts those characters.

Applies to victron-virtual, victron-virtual-switch, and victron-virtual-indicator nodes.

Closes https://github.com/victronenergy/node-red-contrib-victron/issues/525

### DIFF
--- a/src/nodes/victron-virtual-dbus-helpers.js
+++ b/src/nodes/victron-virtual-dbus-helpers.js
@@ -2,6 +2,15 @@ const { addSettings } = require('dbus-victron-virtual')
 const debug = require('debug')('victron-virtual')
 
 /**
+ * Replaces any character that is not alphanumeric or underscore with '_',
+ * producing a string safe for use in D-Bus object path elements and service names.
+ * The original node.id is preserved for filesystem persistence.
+ */
+function sanitizeIdForDbus (id) {
+  return id.replace(/[^A-Za-z0-9_]/g, '_')
+}
+
+/**
  * Parses NODE_RED_DBUS_ADDRESS and returns a tcp: bus address string,
  * or null if not set (caller should fall back to session/system bus).
  */
@@ -125,4 +134,4 @@ function createDebouncedSetters (node, debounce, delayMs) {
   return { shouldApplyImmediately, getDebouncedSetter }
 }
 
-module.exports = { getTcpBusAddress, callAddSettingsWithRetry, getDeviceInstance, registerInputHandler, flushPendingInputs, createDebouncedSetters }
+module.exports = { sanitizeIdForDbus, getTcpBusAddress, callAddSettingsWithRetry, getDeviceInstance, registerInputHandler, flushPendingInputs, createDebouncedSetters }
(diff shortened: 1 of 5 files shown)


src/nodes/victron-virtual-dbus-helpers.js
```diff
@@ -1,6 +1,17 @@
...
 
+/**
+ * Replaces hyphens with underscores in a node ID for use in D-Bus paths.
+ * Node-RED subflow node IDs are formed as '${instanceId}-${templateId}',
+ * and the Venus OS settings daemon rejects hyphens in settings path elements.
+ * Dots (present in old-format Node-RED IDs) are left intact as localsettings
+ * accepts them.
+ */
+function sanitizeIdForDbus (id) {
+  return id.replace(/-/g, '_')
```
Chris927: This contradicts the statement in the PR:

> Replace any non-alphanumeric character with _ before [...]

... but the characters of an id are probably only ascii (plus `.`, as per one of the tests) anyway, other than `-`, so the implementation is probably fine, just the PR description is not.

In case we do want to be pedantic, we may want to cater for cases with other characters, and repeated non-alphanumeric characters. Then this would work:

```js
> 'bla-blu-?ö-test.123'.replace(/[^a-zA-Z0-9\.]+/g, '_')
'bla_blu_test
```

src/nodes/victron-virtual-indicator.js
Chris927, line 152: Is it fine to have the un-sanitized id as value for `Serial`? Those looking at the virtual device via dbus might find it confusing if the Serial is not identical to the id in the service name.

Chris927 (dismissed):
Requesting changes, to be cautious: Depending on the answers to my questions, no changes may actually be required.

I tested this manually, it's working well, but the usefulness of having virtual devices in sub-flows is probably limited: The image below is a virtual indicator in a sub-flow, used twice (in a flow); there is no way to distinguish the two, as labels and name etc. are identical.

Image

dirkjanfaber:
> I tested this manually, it's working well, but the usefulness of having virtual devices in sub-flows is probably limited: The image below is a virtual indicator in a sub-flow, used twice (in a flow); there is no way to distinguish the two, as labels and name etc. are identical.

There are some tricks you can do to work around that. E.g. by injecting something like ` { "/GenericInput/0/Settings/CustomName": $env('customname') }` to get the name from the edit properties of the subflow.

src/nodes/victron-virtual-dbus-helpers.js
dirkjanfaber: Yes, better be strict now. We can always allow extra characters when that is requested. So happy to take your suggestion.

src/nodes/victron-virtual-indicator.js
dirkjanfaber: Lets also sanitise that. I'll update the PR.

Chris927 (approved):

PR #530, opened 2026-06-29
fix(gps): fix D-Bus marshal error for UtcTime on virtual GPS device

### LINKED ISSUE #528: Virtual Device (GPS) bug introduced in 1.7.15 - debounced value issue
**Describe the bug**
The Virtual Device node (GPS type) throws an `AssertionError` when attempting to write values to D-Bus via `dbus-native-victron`. The error `variant data should be [signature, data]` occurs in `marshall.js` when the node attempts to apply a debounced value for a GPS path such as `Speed`. This regression was introduced in v1.7.15 — downgrading to v1.7.14 resolves the issue.

**To Reproduce**
1. Add a Virtual Device node configured as GPS type
2. Send a payload of `{ "Speed": 0.1 }` to the node (a plain object with a D-Bus path key and numeric value, as documented in the node's own UI help text)
3. Observe error: `Failed to apply debounced value for Speed: variant data should be [signature, data]`

**Expected behavior**
The numeric value should be marshalled correctly to D-Bus as a variant double, as it was in v1.7.14.

**Screenshots**
N/A

**Flow**
```json
[SignalK input] → [Function node: msg.payload = { "Speed": msg.payload }] → [Virtual Device node (GPS)]
```

**Hardware**
- Hardware: Windows PC running SignalK server (not Cerbo/Venus OS hardware — Node-RED running via SignalK's embedded Node-RED instance)

**Software**
- node-red-contrib-victron version: 1.7.15 (bug), 1.7.14 (working)
- Venus OS version: N/A — running on SignalK/Windows

**Additional context**
The full error stack:
```
AssertionError [ERR_ASSERTION]: variant data should be [signature, data]
    at write (c:\signalk\signalkhome\.signalk\red\node_modules\dbus-native-victron\lib\marshall.js:71:14)
    at writeStruct (...)
    at marshall (...)
    at Object.marshallMessage [as marshall] (...)
    at EventEmitter.self.message (...)
{
  code: 'ERR_ASSERTION',
  actual: undefined,
  expected: 2,
  operator: '=='
}
```
The GitHub compare between v1.7.14 and v1.7.15 should identify the commit that changed variant handling in the Virtual Device node: https://github.com/victronenergy/node-red-contrib-victron/compare/v1.7.14...v1.7.15

dirkjanfaber:
Virtual GPS nodes threw 'variant data should be [signature, data]' when sending any GPS data field (Speed, Latitude, etc.) because the onPropertiesChanged callback auto-injected UtcTime and dbus-victron-virtual's wrapValue did not handle the 'u' (uint32) type, returning a raw number instead of the required ['u', value] variant pair.

Fix: bump dbus-victron-virtual to 0.1.36, which adds wrapValue support for 'u'. Also adds:
- a wrapValue compatibility test on gps.properties.UtcTime
- a contract test that walks all device-type modules and calls wrapValue on every declared property, catching unsupported types at CI time

closes https://github.com/victronenergy/node-red-contrib-victron/issues/528

### DIFF
--- a/test/victron-virtual-device-type-properties.test.js
+++ b/test/victron-virtual-device-type-properties.test.js
@@ -2,20 +2,10 @@
 /* eslint-env jest */
 const fs = require('fs')
 const path = require('path')
-const { __private__: { wrapValue } } = require('dbus-victron-virtual')
 
 const DEVICE_TYPE_DIR = path.join(__dirname, '../src/nodes/victron-virtual/device-type')
 
-const SAMPLE_VALUES = {
-  b: true,
-  d: 1.0,
-  i: 1,
-  s: 'test',
-  u: 1,
-  ad: [1.0],
-  ai: [1],
-  as: ['test']
-}
+const KNOWN_TYPES = ['b', 'd', 'i', 's', 'u', 'ad', 'ai', 'as']
 
 // Flatten possibly-nested property maps (e.g. generator.js uses { genset: {...}, dcgenset: {...} })
 function flattenProperties (properties) {
@@ -32,7 +22,7 @@ function flattenProperties (properties) {
 
 const deviceModuleFiles = fs.readdirSync(DEVICE_TYPE_DIR).filter(f => f.endsWith('.js'))
 
-describe('virtual device-type property types are wrappable by dbus-victron-virtual', () => {
+describe('virtual device-type property types are known D-Bus types', () => {
   for (const file of deviceModuleFiles) {
     const mod = require(path.join(DEVICE_TYPE_DIR, file))
     if (!mod.properties) continue
@@ -41,12 +31,8 @@ describe('virtual device-type property types are wrappable by dbus-victron-virtu
 
     describe(file, () => {
       for (const [propName, propDesc] of Object.entries(allProperties)) {
-        test(`${propName} (type '${propDesc.type}') wraps to [signature, value]`, () => {
-          const sampleValue = SAMPLE_VALUES[propDesc.type]
-          expect(sampleValue).toBeDefined()
-          const result = wrapValue(propDesc, sampleValue)
-          expect(Array.isArray(result)).toBe(true)
-          expect(result).toHaveLength(2)
+        test(`${propName} type '${propDesc.type}' is a known D-Bus type`, () => {
+          expect(KNOWN_TYPES).toContain(propDesc.type)
         })
       }
     })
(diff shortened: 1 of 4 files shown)


test/victron-virtual-device-type-properties.test.js
```diff
@@ -0,0 +1,54 @@
...
+  for (const file of deviceModuleFiles) {
+    const mod = require(path.join(DEVICE_TYPE_DIR, file))
+    if (!mod.properties) continue
+
+    const allProperties = flattenProperties(mod.properties)
+
+    describe(file, () => {
+      for (const [propName, propDesc] of Object.entries(allProperties)) {
+        test(`${propName} (type '${propDesc.type}') wraps to [signature, value]`, () => {
+          const sampleValue = SAMPLE_VALUES[propDesc.type]
```
Chris927: This is an odd test: What we do is

(a) check that all properties are of a known type (the keys in `SAMPLE_VALUES` object), and
(b) test the `wrapValue` function, which is from another package.

This is confusing, as not unit testing, and it is more complex and harder to understand for the reader than necessary.

What we should do instead:

1. If the other package does not have sufficient test coverage for `wrapValue`, expand the unit tests there.
2. Have a test here that clearly does (a), with a description / test name that states that, something like this:

```js
describe('virtual device-type property types', () => {
  // iterate over all of them
  test(`${file}.${propName} is of known type ${propDesc.type}`, () => {
    KNOWN_TYPES.includes(propDesc.type)
  })
})
```

test/victron-virtual-gps.test.js
```diff
@@ -9,6 +10,16 @@ describe('gps device module', () => {
     expect(typeof gps.onPropertiesChanged).toBe('function')
   })
 
+  describe('UtcTime property', () => {
```
Chris927, line 6: Same feedback as with the other test. A simple `expect(gps.properties.UtcTime.type).toBe('u')` will do as unit test, if that's really necessary.

Chris927 (dismissed):
It's great that we add unit tests, but they are also code, so we should keep them clean and DRY. Therefore requesting changes.

test/victron-virtual-device-type-properties.test.js
dirkjanfaber: Yes, I agree that the test is odd and very doubtful. The (failed) idea was to prevent future wrapping of not yet supported values. But this clearly isn't the way.

Chris927 (approved):

PR #533, opened 2026-06-30
feat(acload): auto-calculate Ac/Energy/Forward from power

dirkjanfaber:
Mirrors the pvinverter auto-energy feature for the virtual AC load. When enabled, integrates power over time (per-phase or total Ac/Power fallback) to accumulate Ac/Energy/Forward. Absent field treats as disabled for safe migration of existing flows.

Extracts accumulateDelta and WATT_MILLISECONDS_PER_KWH into a shared energy-utils.js module to avoid duplication between pvinverter and acload.

### DIFF
--- a/src/nodes/victron-virtual/device-type/pvinverter.js
+++ b/src/nodes/victron-virtual/device-type/pvinverter.js
@@ -93,7 +93,7 @@ function onPropertiesChanged ({ changes, instance, config }) {
     const energyKey = `Ac/L${i}/Energy/Forward`
     const tsKey = `_lastL${i}PowerTimestamp`
     if (powerKey in changes) {
-      accumulateDelta(changes, instance, energyKey, instance[powerKey], instance[tsKey], now)
+      accumulateDelta({ changes, instance, energyKey, oldPower: instance[powerKey], lastTs: instance[tsKey], now })
       instance[tsKey] = now
       anyPhaseUpdated = true
     }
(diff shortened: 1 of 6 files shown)


src/nodes/victron-virtual/device-type/pvinverter.js
```diff
@@ -80,15 +80,6 @@ function initialize (config, ifaceDesc, iface, node) {
   return `Virtual ${iface.NrOfPhases}-phase pvinverter`
 }
 
-function accumulateDelta (changes, instance, energyKey, oldPower, lastTs, now) {
```
Chris927, line 83: [not blocking] As this function has so many parameters, I suggest using named parameters:

```js
function accumulateDelta({ changes, instance, energyKey, oldPower, lastTs, now }) {
  ...
}
```

This makes it easier to have default values, and adjust (add or remove) parameters in future, and it avoids bugs based on "mis-counting" the parameter position.

test/victron-virtual-acload.test.js
```diff
@@ -0,0 +1,296 @@
...
+      const before = Date.now()
+      acload.onPropertiesChanged({
+        changes: { 'Ac/Power': 1000 },
+        instance,
+        config: { acload_auto_energy: true, acload_nrofphases: 1 }
+      })
+      expect(instance._lastPowerTimestamp).toBeGreaterThanOrEqual(before)
+    })
+
+    test('accumulates energy on second power reading', () => {
```
Chris927: [not blocking] We should unit-test `accumulateDelta()` in its own file `energy-utils.test.js`, instead of testing the accumulation here.

Same comment applies to some of the tests below.

Chris927 (approved):
Nothing blocking, but we should move the unit tests to the right place / refactor them to actually be unit tests for the unit under test.

PR #534, opened 2026-07-02
feat(virtual): generalize S2 protocol support beyond acload

dirkjanfaber:
Extract the S2 transport (CEM Connect/Disconnect/Message/KeepAlive, power-measurement broadcast) into a shared enableS2Support() helper so any device-type module can opt in via a supportsS2 export instead of the editor/backend hardcoding the 'acload' device name. Device-type modules own their own S2 resource-settings shape (e.g. acload's on/off hysteresis thresholds) and pass it into the shared transport, so future consumer/producer/storage resource types can define their own shape without touching shared code.

Behavior for existing acload flows is unchanged.

### DIFF
--- a/src/nodes/victron-virtual-browser.js
+++ b/src/nodes/victron-virtual-browser.js
@@ -18,11 +18,18 @@ import {
   getShowUIValue,
   getVirtualNodeLabel,
   determineOutputLabel,
-  setDeviceCapabilities,
-  getDeviceCapabilities,
   fetchDeviceCapabilities
 } from './victron-virtual-functions.js'
 import { initializeTooltips } from './victron-common.js'
+
+// Node-RED can call a node's outputLabels(index) (-> determineOutputLabel) to draw port
+// tooltips for any node already on the canvas, independent of whether its edit dialog has
+// ever been opened. That callback only receives an index, so there is no call site through
+// which fresh capability data could be threaded in - it must be readable synchronously from
+// somewhere. This object is that one place; every other call site (oneditprepare,
+// checkSelectedVirtualDevice, updateOutputs) fetches its own copy and threads it through
+// explicitly instead of reading from here.
+const canvasDeviceCapabilities = {}
 
 window.__victron = {
   checkGeneratorType,
@@ -43,19 +50,18 @@ window.__victron = {
   getShowUIValue,
   initializeTooltips,
   getVirtualNodeLabel,
-  determineOutputLabel,
-  setDeviceCapabilities,
-  getDeviceCapabilities,
+  determineOutputLabel: (node, index) => determineOutputLabel(node, index, canvasDeviceCapabilities),
   fetchDeviceCapabilities
 }
 
-// Node-RED can call a node's outputLabels(index) (-> determineOutputLabel) to draw port
-// tooltips for any node already on the canvas, independent of whether its edit dialog has
-// ever been opened - so warm the device-capability cache as soon as this script loads rather
-// than waiting for oneditprepare. If this hasn't resolved yet when a label/output-count is
-// requested, calculateOutputs/determineOutputLabel fall back to their non-S2 defaults and
-// self-correct on the next render once the fetch completes.
+// Warm canvasDeviceCapabilities as soon as this script loads rather than waiting for
+// oneditprepare, so outputLabels() has S2 capability data available for nodes whose dialog
+// hasn't been opened this session. If this hasn't resolved yet when a label is requested,
+// determineOutputLabel falls back to its non-S2 default and self-corrects on the next render
+// once the fetch completes.
 try {
   const baseUrl = (window.RED?.settings?.httpNodeRoot || window.RED?.settings?.httpAdminRoot || '').replace(/\/$/, '')
-  fetchDeviceCapabilities(baseUrl).then(setDeviceCapabilities).catch(() => {})
-} catch (e) { /* RED not ready yet; oneditprepare's fetch will populate the cache */ }
+  fetchDeviceCapabilities(baseUrl)
+    .then(list => { (list || []).forEach(dt => { canvasDeviceCapabilities[dt.value] = dt }) })
+    .catch(() => {})
+} catch (e) { /* RED not ready yet; outputLabels falls back to non-S2 defaults until this resolves */ }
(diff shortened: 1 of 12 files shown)


src/nodes/victron-virtual-functions.js
```diff
@@ -1040,7 +1062,7 @@ export function updateBatteryVoltageVisibility () {
   $('#battery-voltage-custom-label').toggle(preset === 'custom')
 }
 
-export function checkSelectedVirtualDevice (context) {
+export function checkSelectedVirtualDevice (context, deviceCapabilities = getDeviceCapabilities()) {
```
Chris927: The name of this function is misleading: The name suggests something is checked, which is not the case. The function determines the visibility of form elements, and attaches event listeners to those form elements.

Suggestion: rename to `updateVirtualDeviceFormElements()` or similar.

src/nodes/victron-virtual.html
Chris927, line 111: Why are we not using `fetchDeviceCapabilities()` here? We do the same thing (i.e. to fetch from `/virtual-device-types`, but in two different ways, this should be refactored (to use fetchDeviceCapabilities() here as well).

src/nodes/victron-virtual-functions.js
```diff
@@ -993,6 +993,28 @@ export function fetchEvChargers (baseUrl) {
     })
 }
 
+let deviceCapabilitiesCache = {}
```
Chris927: Global, mutable variables are bad.

Suggestion: We encapsulate the device capabilities using a factory function for the cache, then instantiate a single cache under `window.__victron__.capabilitiesCache`. The factory function would look like this:

```js
function createDeviceCapabilitiesCache() {
  let cachedCapabilities = null // mutable state is not accessible from outside this function

  async function update() {
    const capabilities = await fetchDeviceCapabilities(baseUrl)
    cachedCapabilities = capabilities
  }
  function get() {
    return cachedCapabilities
  }
  return {
    update,
    get
  }
}
```

In one place (probably `src/nodes/victron-virtual-browser.js`) we can now instantiate the cache and call the update:

```js
  const cache = createDeviceCapabilitiesCache()
  window.__victron__.capabilitiesCache = cache
  cache.update().then(...).catch(...)
```

In other places, we can check if we have a cache, and if not, update it:

```js
const cache = window.__victron__.capabilitiesCache
if (!cache.get()) {
  await cache.update()
}
```

Now the question "where is this global variable possibly mutated, and is it valid to do the mutation or not?" completely disappears. And code becomes testable.

Chris927 (changes requested):
I'm requesting changes because this introduces a mutable global variable, which should be avoided.

Further, as per code coverage reported at the end of `npm run test`, `src/nodes/victron-virtual-functions.js` is not covered at all. I *think* this is misleading, as we do test at least some of the functions there, but the rollup into `test/fixtures/victron-virtual-functions.cjs' obfuscates this. I suggest we look into this further: Ideally, coverage reporting should accurately refer to the files under test.

src/nodes/victron-virtual-functions.js
```diff
@@ -1652,41 +1667,39 @@ const DEVICE_TYPE_TO_NUM_OUTPUTS = {
...
  * Calculate the number of outputs for a virtual device
  * @param {string} device - Device type (e.g., 'battery', 'switch', 'gps')
  * @param {object} config - Device configuration object
+ * @param {object} [deviceCapabilities] - Capability metadata keyed by device value
  * @returns {number} Number of outputs (minimum 1)
  */
-export function calculateOutputs (device, config) {
+export function calculateOutputs (device, config, deviceCapabilities = {}) {
   if (DEVICE_TYPE_TO_NUM_OUTPUTS[device]) {
     return DEVICE_TYPE_TO_NUM_OUTPUTS[device](config)
```
Chris927, line 1682: [non-blocking] Depending wether there is an entry in `DEVICE_TYPE_TO_NUM_OUTPUTS` or not, we have different code paths, meaning the supportsS2 logic cannot apply, if there is an entry.

How about making this one codepath:

```js
let numOutputs = 1; // default
if (DEVICE_TYPE_TO_NUM_OUTPUTS[device]) {
  numOutputs = DEVICE_TYPE_TO_NUM_OUTPUTS[device](config)
}
if (deviceCapabilities[device]?.supportsS2 && config?.enable_s2support) {
  numOutputs += 1 // passthrough + S2 signals
}
return numOutputs
```

src/nodes/victron-virtual-browser.js
```diff
@@ -17,10 +17,20 @@ import {
...
 import { initializeTooltips } from './victron-common.js'
 
+// Node-RED can call a node's outputLabels(index) (-> determineOutputLabel) to draw port
+// tooltips for any node already on the canvas, independent of whether its edit dialog has
+// ever been opened. That callback only receives an index, so there is no call site through
+// which fresh capability data could be threaded in - it must be readable synchronously from
+// somewhere. This object is that one place; every other call site (oneditprepare,
+// checkSelectedVirtualDevice, updateOutputs) fetches its own copy and threads it through
+// explicitly instead of reading from here.
+const canvasDeviceCapabilities = {}
```
Chris927, line 32: [just feedback] Still a global, but read-only, and with a good comment :)

src/nodes/victron-virtual.html
Chris927, line 110: [non-blocking] We duplicate determining `baseUrl` and the fetching, rather:

1) Add `fetchDeviceCapabilitiesInBrowser()` to victron-virtual-functions.js, which determines baseUrl and calls `fetchDeviceCapabilities()`
2) Use this function here, instead of setting `baseUrl` and using `$.getJSON()`

src/nodes/victron-virtual/device-type/acload.js
```diff
@@ -184,4 +110,4 @@ function onPropertiesChanged ({ changes, instance, config }) {
   return changes
 }
 
-module.exports = { properties, initialize, onPropertiesChanged, label: 'AC Load' }
+module.exports = { properties, initialize, onPropertiesChanged, label: 'AC Load', supportsS2: true }
```
Chris927, line 113: [non-blocking] Not ideal that you have to both call `enableS2Support()` *and* set `supportsS2: true`, maybe refactor in future.

Chris927, line 113: [non-blocking] Not ideal that you have to both call `enableS2Support()` *and* set `supportsS2: true`, maybe refactor in future.

Chris927 (approved):
Approving, despite some comments remaining: There will be more changes to S2-related logic, so I suggest to tackle them then.

Concerning testing: I was not able to test S2 functionality itself. I tested the editing manually, and experimentally added S2 support for another device type, which worked.

Issue #544, opened 2026-08-04
[Feature] add /Position to virtual acload


dirkjanfaber:
In order to set the position of an AC load, the `/Position` path needs to be added. 

Also check the regular nodes to see if that has support for reading (and setting) the position.

See https://github.com/victronenergy/venus/wiki/dbus#grid-and-acload-and-genset-meter

PR #545, opened 2026-08-04
Add /Position, /PhaseSetting to virtual AC-load-style devices (#544)

dirkjanfaber:
Adds `/Position` (0=AC output, 1=AC input) to the virtual AC load device and the regular acload node, configurable from the edit dialog. Also adds `/PhaseSetting` for single-phase configs, so `/Ac/L[X]/*` reflects the phase the load is actually wired to instead of always defaulting to L1, plus /IsGenericEnergyMeter=1 to match real driver behavior (e.g. dbus-shelly).

The same `Position/PhaseSetting` support is extended to the energy meter device type's acload/evcharger/heatpump roles for consistency.

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -1842,6 +1842,9 @@
 	  $('#battery-voltage-custom-label').toggle(preset === 'custom');
 	}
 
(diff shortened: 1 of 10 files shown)


Chris927 (approved):
Looking good, did some manual testing which went well.

As discussed with @dirkjanfaber offline, strangely my `npm run test:unit` fails, even though it works for him. I'm still accepting, as CI pipeline would block if they fail there.

PR #550, opened 2026-08-22
fix(virtual-switch): respect httpNodeRoot for the switch name/group cache lookup (#471)

### LINKED ISSUE #471: Configuring a deployed virtual switch node causes errors while trying to get values from the cache
**Describe the bug**

When editing a switch node (in my case, a "basic slider"), I get below errors in the console:

```
GET https://venus.local:1881/victron/cache?filter_by_serial=2cdd7d25339e162f 404 (Not Found)
fetchSwitchNodeNameAndGroupFromCache @ victron-virtual-functions.js:1159
oneditprepare @ VM65:91
t @ red.min.js?v=e3326398ab1d:18
k @ red.min.js?v=e3326398ab1d:18
open @ red.min.js?v=e3326398ab1d:18
n @ red.min.js?v=e3326398ab1d:19
show @ red.min.js?v=e3326398ab1d:19
edit @ red.min.js?v=e3326398ab1d:18
mt @ red.min.js?v=e3326398ab1d:18
i @ vendor.js?v=e3326398ab1d:28Understand this error
installHook.js:1 Error fetching name and group from cache for switch node: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```

**To Reproduce**

1. Using node  'Virtual Switch'
2. Choose switch type "Basic Slider", the rest of the configuration can be anything.
3. Deploy
4. Double-click the node
5. The above error occurs on the developer console in the browser.

**Expected behavior**

There should be no error.

**Screenshots**

Image

**Hardware:**
 - Hardware: Cerbo GX, with 

**Software:**
 - node-red-contrib-victron version: node-red-contrib-victron@ee8005e (basically current master, v1.7.10)
 - Venus OS version: not relevant
 
**Additional context**

The likely reason is that we have a hard-coded url [here](https://github.com/victronenergy/node-red-contrib-victron/blob/829dc87695a17d3bf99293e63db676e1eb3b2835/src/nodes/victron-virtual-functions.js#L896), which breaks when `httpNodeRoot` is set (as is the case in my environment: `httpNodeRoot: '/local-dev/'`

ryanmarshall:
## Problem
If you set `httpNodeRoot` in Node-RED, opening a deployed Virtual Switch node
shows an error in the browser console:

    GET .../victron/cache?filter_by_serial=… 404 (Not Found)
    Error fetching name and group from cache for switch node:
      SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON

The switch node asks for `/victron/cache`, but that path is wrong when
`httpNodeRoot` is set. The cache route is added to `RED.httpNode`, so Node-RED
serves it under `httpNodeRoot` (for example `/local-dev/victron/cache`). The old
code left the prefix off, so the request hit the wrong path and got a 404. The
404 page is HTML, and the code then tried to read that HTML as JSON, which is
what throws the error.

Two other functions in the same file (`fetchEvChargers` and
`fetchDeviceCapabilities`) already add the prefix. This one did not.

## Fix
Add the `httpNodeRoot` prefix to the request, the same way those two functions
do. It is one line in the function and one line in the caller
(`victron-virtual-switch.html`).

## Tests
Added tests to `victron-virtual-functions.switch.test.js`:
- The request now includes the prefix, and never uses the old wrong path.
- With no prefix set, the path stays the same as before (nothing else changes).
- A 404 HTML response still produces the same error from the bug report, so the
  failure is covered.

`npm test` passes (1151 tests).

## Verified
Tested on a local Node-RED with `httpNodeRoot: '/local-dev/'`.

Before — the request goes to the wrong path and fails:

image-1787362700056

After — the request goes to the correct path and reaches the endpoint:

    GET /local-dev/victron/cache?filter_by_serial=…  ->  reaches the endpoint

(This test machine has no GX connected, so the endpoint returns 503. A real GX
returns the cache as JSON.)

Fixes #471

### DIFF
--- a/resources/victron-virtual-functions.js
+++ b/resources/victron-virtual-functions.js
@@ -1795,12 +1795,12 @@
 	    .then(response => response.json())
 	}
 
-	function fetchSwitchNodeNameAndGroupFromCache (id) {
+	function fetchSwitchNodeNameAndGroupFromCache (id, baseUrl) {
 	  if (!id) {
 	    return Promise.reject(new Error('id is required'))
 	  }
 
-	  return fetch(`/victron/cache?filter_by_serial=${id}`)
+	  return fetch(`${baseUrl || ''}/victron/cache?filter_by_serial=${id}`)
 	    .then(response => response.json())
 	    .then(data => {
 	      for (const key in data) {
--- a/src/nodes/victron-virtual-functions.js
+++ b/src/nodes/victron-virtual-functions.js
@@ -1004,12 +1004,12 @@ export function fetchDeviceCapabilities (baseUrl) {
     .then(response => response.json())
 }
 
-export function fetchSwitchNodeNameAndGroupFromCache (id) {
+export function fetchSwitchNodeNameAndGroupFromCache (id, baseUrl) {
   if (!id) {
     return Promise.reject(new Error('id is required'))
   }
 
-  return fetch(`/victron/cache?filter_by_serial=${id}`)
+  return fetch(`${baseUrl || ''}/victron/cache?filter_by_serial=${id}`)
     .then(response => response.json())
     .then(data => {
       for (const key in data) {
--- a/src/nodes/victron-virtual-switch.html
+++ b/src/nodes/victron-virtual-switch.html
@@ -97,7 +97,8 @@
                 initializeTooltips();
             }
 
-            fetchSwitchNodeNameAndGroupFromCache(this.id).then(({name, group}) => {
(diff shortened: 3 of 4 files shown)


dirkjanfaber (approved):
Verified the fix: `/victron/cache` is registered on `RED.httpNode` (config-client.js), so `httpNodeRoot` prefixing is correct and matches the existing pattern used by `fetchEvChargers`/`fetchDeviceCapabilities`. Rebuilt and confirmed `resources/victron-virtual-functions.js` matches the source with no drift. Full suite passes (1151/1151), including the new regression tests for #471.

PR #551, opened 2026-08-24
feat(virtual): streamline meter device selection, retire Energy meter picker

dirkjanfaber:
Consolidates AC Load, Generator, Heat pump, and EV charger virtual devices onto a shared minimal-meter D-Bus shape, replacing the old "Energy meter" node's Role dropdown (kept registered internally so already-deployed flows keep working).

- Add standalone Heat pump and EV charger device types.
- AC Load, Generator (AC), and Heat pump gain a "use as grid meter only" checkbox for reduced/grid-meter-style reporting. All three (and EV charger) stay registered under their own D-Bus service name, with a productType override so the reported ProductId still reflects the grid meter when used as one.
- Remove AC Load's unused `S2/0/RmSettings/*` properties.
- Correct `IsGenericEnergyMeter` and `Ac/PowerFactor` defaults to match the device's actual role/mode (affects AC Load, Heat pump, and the legacy Energy meter node).
- Add "Usage" doc entries for Heat pump/EV charger; misc tooltip wording fixes.

### DIFF
--- a/src/nodes/victron-virtual.html
+++ b/src/nodes/victron-virtual.html
@@ -122,6 +122,11 @@
                     const $select = $('#node-input-device');
                     $select.empty();
                     deviceTypes.forEach(function(dt) {
+                        // Legacy-only types (e.g. retired "Energy meter") aren't offered for new
+                        // nodes, but still need to appear as an option so an existing node using
+                        // one resolves/displays correctly instead of silently falling back to
+                        // whatever option happens to be first in the list.
+                        if (dt.legacyOnly && dt.value !== self.device) return;
                         $select.append($('<option>', { value: dt.value, text: dt.label }));
                         deviceCapabilities[dt.value] = dt;
                     });
--- a/src/nodes/victron-virtual.html
+++ b/src/nodes/victron-virtual.html
@@ -14,8 +14,17 @@
             acload_nrofphases: {value: 1, required: false},
             acload_phasesetting: {value: 1, required: false},
             acload_auto_energy: {value: true, required: false},
+            acload_grid_meter_only: {value: false, required: false},
             enable_s2support: {value: false, required: false},
             s2_measurement_type: {value: '3_PHASE_SYMMETRIC', required: false},
+            // heatpump
+            heatpump_position: {value: 0, required: false},
+            heatpump_nrofphases: {value: 1, required: false},
+            heatpump_phasesetting: {value: 1, required: false},
+            heatpump_auto_energy: {value: true, required: false},
+            heatpump_grid_meter_only: {value: false, required: false},
+            // evcs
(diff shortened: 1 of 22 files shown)


src/nodes/victron-virtual.html
```diff
@@ -117,6 +120,11 @@
                     const $select = $('#node-input-device');
                     $select.empty();
                     deviceTypes.forEach(function(dt) {
+                        // Legacy-only types (e.g. retired "Energy meter") aren't offered for new
+                        // nodes, but still need to appear as an option so an existing node using
+                        // one resolves/displays correctly instead of silently falling back to
+                        // whatever option happens to be first in the list.
+                        if (dt.legacyOnly && dt.value !== self.device) return;
```
Chris927: For discussion: Why not make the type dropdown read-only, once a node has been deployed? Changing the device type of an already-deployed virtual device likely causes undesired / difficult-to-understand consequences (e.g. loss (or not?) of persisted data, ...). A physical device cannot change its type, if you use a device of a different (physical) type you have to remove the node and add another one; we could do the same with virtual devices.

This would simplify handling "legacy" virtual device types: If the type is read-only, it just stays there (no "select" input needed), legacy or not. If the type is not read-only yet (because the virtual device had never been deployed), then we show all non-legacy types.

This might also make it easier in future to deprecate virtual device types and add new ones (we could have an "EV charger V2" which supersedes the current one, etc.

src/nodes/victron-virtual.html
```diff
@@ -734,8 +754,13 @@
 The AC Load device can be configured by selecting the number of phases (1 or 3)
 that are available on the device.
 
+By default it presents as a plain generic energy meter (minimal measurements
+only, no Position). Enabling S2 support is what turns it into its own full
```
Chris927, line 766: "Enabling S2 ... its own full AC Load device". Really? It requires a bit more, as I understand it. This sounds like an advertisment, not technical documentation.

Chris927 (commented):
Comments for discussion, I've barely tested manually yet, will do so now.

src/nodes/victron-virtual.html
dirkjanfaber: While I agree that it sounds like an ad, I can't think of a much better way to phrase it. The S2 support adds control abilities to the, until that moment, energy metering device.

src/nodes/victron-virtual.html
dirkjanfaber: This is a valid one and I agree.

Chris927 (approved):
Reviewed, and tested manually, and discussed a few migration challenges with @dirkjanfaber . Also raised #552 to not have this gap in the energy calculations block this PR.

PR #556, opened 2026-09-02
fix(services):  fill dbus path gaps (and repair check-services-json.js)

dirkjanfaber:
The audit script still assumed the retired input-X/output-X schema and crashed immediately against the current flat services.json. Rewrote it to compare against dbus_modbustcp's attributes.csv (no more unused exceljs dependency) and to expand wildcard path templates (e.g. /Ac/Out/{phase}/V) before diffing, fixing a large false-positive count. Wired it in as `npm run check-services`.

While auditing, found and fixed a stray `}` in acsystem's /Ac/Out/{phase}}/V that broke its wildcard expansion, and added several genuinely missing dbus paths surfaced by the fixed script: temperature /IAQS, battery alarm codes, DESS status fields, vebus active-input current limit, solarcharger PV current and per-tracker MPP mode, acsystem PV disable, and DC genset status/error/product fields to match their AC genset counterparts.

Regenerated src/nodes/config-client.html via `npm run documentation`.

### DIFF
--- a/scripts/check-services-json.js
+++ b/scripts/check-services-json.js
@@ -1,197 +1,140 @@
 // Also see https://github.com/victronenergy/node-red-contrib-victron/issues/115
+//
+// Compares services.json against the dbus_modbustcp attributes.csv (see
+// https://github.com/victronenergy/dbus_modbustcp/blob/master/attributes.csv),
+// which lists every dbus path that project knows about together with its
+// type/enum/unit info. That csv is the most complete public reference we
+// have for dbus paths, even though not every dbus path ends up exposed over
+// Modbus TCP (see scripts/exclude.txt for known, intentional gaps).
 
 const fs = require('fs')
-const Excel = require('exceljs')
 
 const argv = require('yargs/yargs')(process.argv.slice(2))
-  .usage('Usage: $0 -s services.json -x modbus_tcp.xlsx')
+  .usage('Usage: $0 -s services.json -c attributes.csv')
   .option('services', {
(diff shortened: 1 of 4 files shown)


scripts/check-services-json.js
```diff
@@ -1,197 +1,168 @@
...
-        if (node.match(/^input-*/) && services['input-relay'][node.replace('input-', '')]) {
-          return
-        }
-        if (node.match(/^output-*/) && services['output-relay'][node.replace('output-', '')]) {
-          return
-        }
-        console.log('// Missing relay service for ' + node + ' ' + path)
-        return
-      }
+const parse = require('csv-parse/lib/sync')
```
Chris927: This fails with the latest version of node, as it seems to prohibit un-exported imports. This fixes it:

```patch
diff --git a/scripts/check-services-json.js b/scripts/check-services-json.js
index d99eba3..bcbd350 100644
--- a/scripts/check-services-json.js
+++ b/scripts/check-services-json.js
@@ -46,7 +46,7 @@ const SKIP_IFACES = new Set([

 const services = JSON.parse(fs.readFileSync(servicesJSON))

-const parse = require('csv-parse/lib/sync')
+const { parse } = require('csv-parse/sync')

 const rows = parse(fs.readFileSync(csvFile, 'utf8'), {
   relax_column_count: true,
```

Chris927 (changes requested):
Needs an import fix, as per comment, otherwise looking good.

PR #560, opened 2026-09-03
fix(nodes): stop removeStatusListener from wiping other nodes' status

dirkjanfaber:
statusListeners.filter(o => o.id === id) kept only the listener being removed and discarded every other node's entry. Since removeStatusListener runs on every input node's close event (fired for touched nodes on any deploy, not just full deploys), other Custom Input nodes silently lost their status tracking and their editor status froze at whatever it last was - typically the green "Connected" dot, even after the underlying D-Bus service (e.g. a Shelly switch/relay) had disappeared.

Confirmed against a real Shelly heatpump device: unplugging it fully removes its D-Bus service (NameOwnerChanged), so the SERVICE_REMOVE event was already being emitted correctly - it just never reached other nodes.

### DIFF
--- a/src/nodes/config-client.js
+++ b/src/nodes/config-client.js
@@ -138,7 +138,7 @@ module.exports = function (RED) {
       return id
     }
 
-    this.removeStatusListener = id => { statusListeners = statusListeners.filter(o => o.id === id) }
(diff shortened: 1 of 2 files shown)


Chris927 (approved):
Such a good find!

PR #568, opened 2026-09-07
fix(virtual): make acload/heatpump Position/PhaseSetting configurable without S2

### LINKED ISSUE #567: Set position for virtual acload, not only when S2 enabled
At the moment you can only set the position for the virtual AC load when S2 is enabled. It also needs to be configurable when S2 is disabled.

(See f2472fd2d11bc277c1ba0978b2012ade4d674516)

dirkjanfaber:
Position and PhaseSetting were only added when enable_s2support was true, even though they're meaningful regardless of S2 support (the energy meter device type already treats them independently of any S2 concept). S2 support still separately gates only the S2 transport paths and whether the device reports as a full acload/heatpump vs. a generic energy meter.

Closes #567

### DIFF
--- a/src/services/dbus-listener.js
+++ b/src/services/dbus-listener.js
@@ -241,7 +241,7 @@ class VictronDbusListener {
           if (res) {
             const deviceInstance = res[1]?.[0]
             if (deviceInstance === undefined) {
-              console.error(`deviceInstance could not be assigned because res[1][0] is undefined owner=${owner} services[owner]=${JSON.stringify(this.services[owner])} (${this.services[owner].name})`)
+              console.error(`deviceInstance could not be assigned because res[1][0] is undefined owner=${owner} services[owner]=${JSON.stringify(this.services[owner])}`)
             }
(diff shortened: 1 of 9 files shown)


src/services/dbus-listener.js
```diff
@@ -250,6 +250,12 @@ class VictronDbusListener {
       }
     })
 
+    // the owner may have disconnected (NameOwnerChanged) while GetValue was in flight,
+    // in which case the entry was already removed from this.services
+    if (!this.services[owner]) {
```
dirkjanfaber, line 255: Add a `console.warn` here.

src/services/dbus-listener.js
```diff
@@ -250,6 +250,12 @@ class VictronDbusListener {
...
+      return
```
Chris927, line 257: We should have a `console.warn()` in this case.

Chris927 (changes requested):
As per our discussion, please add the `console.warn()`. Other than that, I've tested manually successfully.

PR #571, opened 2026-09-08
victron-virtual: declare energymeter_role / energymeter_position in editor defaults (fixes #570)

### LINKED ISSUE #570: victron-virtual energymeter: energymeter_role / energymeter_position read by runtime but not declared in editor defaults — silently reverts to a grid meter on any editor deploy
## Summary
`victron-virtual` (device type `energymeter`) reads `config.energymeter_role` and `config.energymeter_position` at runtime, but neither property is declared in the editor's `defaults` in `src/nodes/victron-virtual.html`. Because the Node-RED editor only retains properties listed in `defaults` when it loads a flow, these two values are silently dropped on **any** editor deploy — even a deploy of an unrelated node on another tab. The node then falls back to the default role and re-registers as a **grid meter**.

Confirmed on **1.7.24** (installed) and by inspection of **1.7.26** (`master`/`v1.7.26`) — the defaults are still absent there.

## Where
- Runtime reads them: `src/nodes/victron-virtual/device-type/energymeter.js`
  - `return ROLE_TO_SERVICE_TYPE[config.energymeter_role] || 'grid'` (service type selection)
  - `iface.Position = Number(config.energymeter_position ?? 0)`
- Editor does **not** declare them: `src/nodes/victron-virtual.html` → `defaults: { … }` has sections for acload / heatpump / evcs / grid / pvinverter / switch / tank / ev / … but no `energymeter_*` keys.

## Impact
With `energymeter_role: "evcharger"` set (via the admin API, since there is no UI field either), the device registers as `com.victronenergy.evcharger.*` and works correctly. After the next editor deploy the role is lost, the device re-registers as `com.victronenergy.grid.*`, and:
1. `dbus-systemcalc-py` treats it as an AC-input grid meter and adds its full power to **AC Consumption** — for a charger physically on the inverter output this **double-counts the load** (we measured ~15 kW reported vs ~8 kW real during a 7 kW EV charge).
2. The EV charger disappears from VRM (the EVCS widget goes blank).

Because the loss happens at editor load time, "modified nodes only" deploys do not help — the editor saves the full flow from its in-memory copy.

## Reproduce
1. Create a `victron-virtual` node, device `energymeter`.
2. Via the admin API, set `energymeter_role: "evcharger"` on it and POST the flow. Device registers as `evcharger`.
3. Open the editor, change any unrelated node, Deploy.
4. Inspect the flow: `energymeter_role` is gone; on the GX, `dbus -y` now lists `com.victronenergy.grid.virtual_<id>`.

## Testing we did
- Simulated the editor's import (keep core fields + declared `defaults` keys) against the unpatched html: the role is stripped — matches the field behaviour above.
- Same simulation against the html with the two keys declared: the role is kept and the node is byte-identical to the live flow.
- Applied the declaration, restarted Node-RED (node html is read at startup — the served `GET /nodes` html did not include the keys until restart), then performed a real editor-style deploy: role retained, device stayed `evcharger`, AC Consumption correct.

## Suggested fix
Declare the properties in `victron-virtual.html` `defaults` (an empty role still resolves to `'grid'` at runtime, so existing nodes are unaffected):
```js
// energymeter
energymeter_role: {value: "", required: false},
energymeter_position: {value: 0, required: false},
```
Ideally also expose them in the edit dialog for the `energymeter` device type, since the runtime already supports the role/position but there is currently no way to set them from the UI.

svwhisper:
## Problem
`victron-virtual` (device type `energymeter`) reads `config.energymeter_role` and `config.energymeter_position` at runtime (`src/nodes/victron-virtual/device-type/energymeter.js`), but neither is declared in the editor's `defaults` in `src/nodes/victron-virtual.html`. The Node-RED editor only keeps declared properties when it loads a flow, so **any** editor deploy — even of an unrelated node — silently drops both values. The device then falls back to the default `grid` role, re-registers as `com.victronenergy.grid.*`, and for a charger on the inverter output its power is double-counted in dbus-systemcalc's AC Consumption while the EVCS disappears from VRM. Details and reproduction in #570.

## Fix
Declare the two properties in `defaults`, in the same style as the other device-type sections. An empty `energymeter_role` still resolves to `'grid'` at runtime (`ROLE_TO_SERVICE_TYPE[...] || 'grid'`), so existing nodes behave exactly as before.

## Testing
- Simulated the editor's import (core fields + declared `defaults` keys): with the unpatched html the role is stripped (reproducing the field failure); with this change it is retained and the node round-trips byte-identical.
- Applied the change to a live installation (1.7.24), restarted Node-RED (node html is read at startup), and confirmed in the running editor that `RED.nodes.getType('victron-virtual').defaults` contains both keys and the editor's post-import copy of the node keeps `energymeter_role: "evcharger"`.
- Live regression/restore test: stripping the role reproduced the double-count (AC Consumption ≈ inverter output + the charger's 7.4 kW again); restoring it removed it; with this patch in place a real editor-style deploy no longer loses the role.

A follow-up could expose these two fields in the edit dialog for the `energymeter` device type, since the runtime already supports them but there is currently no UI to set them.

Fixes #570

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/nodes/victron-virtual.html
+++ b/src/nodes/victron-virtual.html
@@ -40,6 +40,8 @@
             // energymeter
             energymeter_role: {value: "", required: false},
             energymeter_position: {value: 0, required: false},
+            energymeter_nrofphases: {value: 1, required: false},
+            energymeter_phasesetting: {value: 1, required: false},
             // motordrive
             include_motor_temp: {value: false},
             include_controller_temp: {value: false},
--- a/src/nodes/victron-virtual.html
+++ b/src/nodes/victron-virtual.html
@@ -37,6 +37,9 @@
             generator_grid_meter_only: {value: false, required: false},
             // grid
             grid_nrofphases: {value: 1, required: false},
+            // energymeter
+            energymeter_role: {value: "", required: false},
+            energymeter_position: {value: 0, required: false},
             // motordrive
             include_motor_temp: {value: false},
             include_controller_temp: {value: false},
--- a/test/victron-virtual-energymeter-defaults.test.js
+++ b/test/victron-virtual-energymeter-defaults.test.js
@@ -1,0 +1,48 @@
+/* eslint-env jest */
+/**
+ * Regression test for issue #570.
+ *
+ * The Node-RED editor only retains a node's properties across a deploy if
+ * they are declared in RED.nodes.registerType's `defaults` object (in the
+ * node's .html file). Any config key read by a device-type module at
+ * runtime but missing from `defaults` is silently dropped on the next
+ * editor deploy - even a deploy of an unrelated node.
+ *
+ * This asserts every `config.energymeter_*` key read by
+ * device-type/energymeter.js is declared in victron-virtual.html's
+ * `defaults`, so this class of bug can't silently reappear for this
+ * device type.
+ */
+
+const fs = require('fs')
+const path = require('path')
+
+const energymeterSource = fs.readFileSync(
+  path.join(__dirname, '..', 'src', 'nodes', 'victron-virtual', 'device-type', 'energymeter.js'),
+  'utf8'
+)
+const htmlSource = fs.readFileSync(
+  path.join(__dirname, '..', 'src', 'nodes', 'victron-virtual.html'),
+  'utf8'
+)
+
+const configKeys = [...new Set(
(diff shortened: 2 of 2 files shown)


dirkjanfaber (approved):
Good find. Added the other missing fields too.

